PR #23, opened 2024-08-07
Remove id from TestSuite as its not specified in schema

brian-mcnamara:
As per https://windyroad.com.au/dl/Open%20Source/JUnit.xsd (referenced as part of https://bazel.build/reference/test-encyclopedia `XML_OUTPUT_FILE`), I do not see id as an attribute for testsuite

### DIFF
--- a/server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/TestXmlParser.kt
+++ b/server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bep/TestXmlParser.kt
@@ -38,8 +38,6 @@ data class TestSuite(
   val errors: Int,
   @JacksonXmlProperty(isAttribute = true)
   val time: Double,
-  @JacksonXmlProperty(isAttribute = true)
-  val id: Int,
   @JacksonXmlElementWrapper(useWrapping = false)
   @JacksonXmlProperty(localName = "testcase")
   val testcase: List<TestCase> = emptyList(),


Otter-man:
please run `bazel run //tools/format:format`

agluszak:
@mielnikk

brian-mcnamara:
> please run `bazel run //tools/format:format`

Rebased, it looks like there was an issue with installPlugins.sh that was resolved in another commit

mielnikk:
Are you sure it's not specified? Because the schema contains this:
```
<xs:attribute name="id" type="xs:int" use="required">
  <xs:annotation>
    <xs:documentation xml:lang="en">Starts at '0' for the first testsuite and is incremented by 1 for each following testsuite</xs:documentation>
  </xs:annotation>
</xs:attribute>
```
I can see the id in generated JUnit4 test results as well.

brian-mcnamara:
> Are you sure it's not specified? Because the schema contains this:
> 
> ```
> <xs:attribute name="id" type="xs:int" use="required">
>   <xs:annotation>
>     <xs:documentation xml:lang="en">Starts at '0' for the first testsuite and is incremented by 1 for each following testsuite</xs:documentation>
>   </xs:annotation>
> </xs:attribute>
> ```
> 
> I can see the id in generated JUnit4 test results as well.

Ah yep, you are right, missed it was an extension of the complexType and thought it was an attribute on testSuites.

PR #56, opened 2024-09-10
[feature] Add .bazelrc highlighting lexer | #BAZEL-1186

mtoader:
This is a basic highlighting lexer for `*.bazelrc` files. 

It handles quoting and new line continuations.


mtoader:
I have some questions @agluszak:
 
 * How can I get some eyes here? 
 * Should I raise a youtrack ticket to track the work?

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/BazelrcLanguage.kt
```diff
@@ -0,0 +1,18 @@
...
+import javax.swing.Icon
+
+object BazelrcLanguage : Language("Bazelrc")
+
+object BazelrcFileType : LanguageFileType(BazelrcLanguage) {
+  override fun getName(): String = "Bazelrc"
+
+  override fun getDescription(): String = "Bazelrc language"
+
+  override fun getDefaultExtension(): String = "blazerc"
```
agluszak, line 15: I guess it should be bazelrc, since Hirschgarten doesn't support Blaze anyway :)

agluszak:
LGTM (besides one nit), I asked another person from the team to have a look at the lexer-y stuff and he said it's all good.

No need for a ticket, unless there's something you think is missing or to be improved. We can merge it tomorrow probably :)

mtoader:
> LGTM (besides one nit), I asked another person from the team to have a look at the lexer-y stuff and he said it's all good.
> 
> No need for a ticket, unless there's something you think is missing or to be improved. We can merge it tomorrow probably :)

I have follow-ups I want (a psi tree, some semantics completion/highlighting, and some validations (e.g., if you set the same variable twice with different values, validation that the options values match their type, etc.). I'm fine with having them trickle in, too. It was more of a question of whether you would like this work tracked.

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/BazelrcLanguage.kt
mtoader: Changed

jastice:
> I have follow-ups I want (a psi tree, some semantics completion/highlighting, and some validations (e.g., if you set the same variable twice with different values, validation that the options values match their type, etc.). I'm fine with having them trickle in, too. It was more of a question of whether you would like this work tracked.

In this case, a ticket, and ticket references in the PR (eg `#BAZEL-12345 fixed`) are appreciated, so that they will automatically be added to the release notes

agluszak (approved):

PR #73, opened 2024-09-20
[feature] Complete config across imports | #BAZEL-1186

mtoader:
Also fix import resolution of sibling file (e.g. import `.xxx.bazelrc`)

### DIFF
--- a/plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/completion/BazelrcConfigDeclarationsProcessor.kt
+++ b/plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/completion/BazelrcConfigDeclarationsProcessor.kt
@@ -29,7 +29,7 @@ class BazelrcConfigDeclarationsProcessor(private val inputElement: BazelrcLine) 
   fun processElement(current: BazelrcElement) =
     when (current) {
       is BazelrcLine -> processLine(current)
-      is org.jetbrains.bazel.languages.bazelrc.psi.BazelrcImport -> processImport(current)
+      is BazelrcImport -> processImport(current)
       else -> {}
     }
 
--- a/plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/completion/BazelrcConfigDeclarationsProcessor.kt
+++ b/plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/completion/BazelrcConfigDeclarationsProcessor.kt
@@ -1,0 +1,45 @@
+package org.jetbrains.bazel.languages.bazelrc.completion
+
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.util.Processor
+import org.jetbrains.bazel.languages.bazelrc.psi.BazelrcElement
+import org.jetbrains.bazel.languages.bazelrc.psi.BazelrcFile
+import org.jetbrains.bazel.languages.bazelrc.psi.BazelrcLine
+import org.jetbrains.kotlin.utils.addToStdlib.ifTrue
+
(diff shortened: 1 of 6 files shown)


mtoader:
@mpiekutowski can i have some eyes on this too when possible ?

mtoader:
@tpasternak can you take a look ?

tpasternak:
Sorry for that, but I was OOO for two weeks and now I'm flooded with housekeeping tasks in he native plugin

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/completion/BazelrcConfigDeclarationsProcessor.kt
```diff
@@ -0,0 +1,45 @@
...
+
+  override fun process(currentElement: BazelrcElement): Boolean {
+    if (currentElement == inputElement) return true
+    processElement(currentElement)
+    return true
+  }
+
+  fun processFile(file: BazelrcFile?) {
+    file?.virtualFile?.apply {
+      seenFiles.add(this).ifTrue {
```
agluszak, line 22: `ifTrue` seems to come from some weird package, could you remove it please?

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/completion/BazelrcConfigDeclarationsProcessor.kt
```diff
@@ -0,0 +1,45 @@
...
+        file.imports.map(::processElement)
+        file.lines.map(::processElement)
+      }
+    }
+  }
+
+  fun processElement(current: BazelrcElement) =
+    when (current) {
+      is BazelrcLine -> processLine(current)
+      is org.jetbrains.bazel.languages.bazelrc.psi.BazelrcImport -> processImport(current)
```
agluszak, line 32:
```suggestion
      is BazelrcImport -> processImport(current)
```

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/completion/BazelrcConfigDeclarationsProcessor.kt
```diff
@@ -0,0 +1,45 @@
...
+      else -> {}
+    }
+
+  private fun processLine(line: BazelrcLine) {
+    (line.containingFile as? BazelrcFile)?.apply {
+      line.configName()?.let { results[it] = this }
+    }
+  }
+
+  private fun processImport(element: org.jetbrains.bazel.languages.bazelrc.psi.BazelrcImport) {
```
agluszak, line 42:
```suggestion
  private fun processImport(element: BazelrcImport) {
```

agluszak (changes requested):

agluszak:
Otherwise LGTM!

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/completion/BazelrcConfigDeclarationsProcessor.kt
mtoader: No idea how this remained. I saw idea doing some half work when moving code around ( non shortinging imports correctly) but this is just one branch of the when .. doesn't really make sense :)

Thanks for catching

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/completion/BazelrcConfigDeclarationsProcessor.kt
mtoader: The package started with org.jetbrains :) that automatically doesn't make it weird :) Fixed.

agluszak (approved):

PR #77, opened 2024-09-21
Improve allModuleDependencies performance

anupc-db:
For modules with a large number of targets where many targets depend on a common subset of targets, allModuleDependencies can have exponential runtime. This happens because duplicate entries can end up in `toResolve` from different sources.

Instead, make sure that we only process each target once.

Added a test that will timeout within 2 minutes if not optimized.

### DIFF
--- a/server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
+++ b/server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
@@ -53,11 +53,17 @@ object DependencyMapper {
     val toResolve = mutableListOf<Label>()
     toResolve.addAll(module.directDependencies)
     val accumulator = HashSet<Library>()
+    val allSeenTargets = toResolve.toMutableSet()
     while (toResolve.isNotEmpty()) {
       val lib = project.libraries[toResolve.removeLast()]
       if (lib != null && !accumulator.contains(lib)) {
         accumulator.add(lib)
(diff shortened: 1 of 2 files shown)


server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
```diff
@@ -53,11 +53,18 @@ object DependencyMapper {
...
       val lib = project.libraries[toResolve.removeLast()]
       if (lib != null && !accumulator.contains(lib)) {
         accumulator.add(lib)
-        toResolve.addAll(lib.dependencies)
+        allSeenTargets.add(lib.label)
+        for (dep in lib.dependencies) {
+          if (!allSeenTargets.contains(dep)) {
+            allSeenTargets.add(dep)
+            toResolve.add(dep)
+          }
```
xuansontrinh, line 66: can just be:
```kotlin
if (allSeenTargets.add(dep)) {
  toResolve.add(dep)
}
```

xuansontrinh (changes requested):

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/DependencyMapper.kt
anupc-db: Done

xuansontrinh (approved):

xuansontrinh:
Thank you for your contribution.

PR #79, opened 2024-09-24
Cache library modules in buildDependencyModules

anupc-db:
`extractMavenDependencyInfo(library)` is a fixed value within a request context. Cache their value across all targets in a request. This makes a substantial improvement when a large number of targets are queried.

### DIFF
--- a/server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/BspProjectMapper.kt
+++ b/server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/BspProjectMapper.kt
@@ -568,21 +568,25 @@ class BspProjectMapper(
 
   fun buildDependencyModules(project: Project, params: DependencyModulesParams): DependencyModulesResult {
     val targetSet = params.targets.toSet()
+    val cache = mutableMapOf<String, List<DependencyModule>>()
     val dependencyModulesItems =
       project.modules.filter { targetSet.contains(BuildTargetIdentifier(it.label.value)) }.map { module ->
         val buildTargetId = BuildTargetIdentifier(module.label.value)
+        val moduleDependencies = DependencyMapper.allModuleDependencies(project, module)
         val moduleItems =
-          DependencyMapper.allModuleDependencies(project, module).flatMap { libraryDep ->
-            if (libraryDep.outputs.isNotEmpty()) {
-              val mavenDependencyModule = DependencyMapper.extractMavenDependencyInfo(libraryDep)
-              val dependencyModule = DependencyModule(libraryDep.label.value, mavenDependencyModule?.version ?: "")
-              if (mavenDependencyModule != null) {
-                dependencyModule.data = mavenDependencyModule
-                dependencyModule.dataKind = DependencyModuleDataKind.MAVEN
+          moduleDependencies.flatMap { libraryDep ->
+            cache.getOrPut(libraryDep.label.value) {
+              if (libraryDep.outputs.isNotEmpty()) {
+                val mavenDependencyModule = DependencyMapper.extractMavenDependencyInfo(libraryDep)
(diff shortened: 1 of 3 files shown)


xuansontrinh:
the change looks good to me.
could you please also attach a small benchmark to indicate the improvement?

anupc-db:
> the change looks good to me. could you please also attach a small benchmark to indicate the improvement?

Let me know if this change makes sense as a benchmark.

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/BspProjectMapper.kt
```diff
@@ -600,4 +579,36 @@ class BspProjectMapper(
 
     return toRustWorkspaceResult(requestedModules, allRustModules)
   }
+
+  companion object {
+    @JvmStatic
+    fun buildDependencyModulesStatic(project: Project, params: DependencyModulesParams): DependencyModulesResult {
+      val targetSet = params.targets.toSet()
+      val cache = mutableMapOf<String, List<DependencyModule>>()
```
gottagofaster236, line 587: I think this cache can be an instance field of `BazelProjectMapper`, then you won't need the static function

gottagofaster236: Other than this the change looks good

anupc-db: I made it a static function because I couldn't figure out how to create a BspProjectMapper instance (had to fake out a bunch of other fields). Any pointers there would be appreciated.

anupc-db: Hang on - does this being a static function mean that `cache` will be shared across all function calls? That was not my intention.

(This is my first foray into Kotlin)

gottagofaster236: Sorry, I just overlooked this, your code makes sense actually

gottagofaster236: I'm gonna fix the failing test in a second

server/server/src/test/kotlin/org/jetbrains/bsp/bazel/server/sync/BspProjectMapperTest.kt
```diff
@@ -0,0 +1,149 @@
...
+import org.jetbrains.bsp.bazel.server.model.SourceSet
+import org.junit.jupiter.api.Test
+import java.net.URI
+import java.nio.file.Paths
+
+class BspProjectMapperTest {
+  private val cacheLocation =
+    "file:///home/user/.cache/bazel/_bazel_user/ae7b7b315151086e31e3b97f9ddba009/execroot/monorepo/bazel-out/k8-fastbuild-ST-4a519fd6d3e4"
+
+  @Test
```
xuansontrinh, line 25: I think you can add@Timeout annotation here to make the expectation more verbose and better documentation, [link for reference](https://junit.org/junit5/docs/5.5.1/api/org/junit/jupiter/api/Timeout.html)

xuansontrinh (changes requested):

server/server/src/test/kotlin/org/jetbrains/bsp/bazel/server/sync/BspProjectMapperTest.kt
anupc-db: Added `@Timeout` annotation. However it doesn't seem to actually timeout at 30s under `bazel test`.

xuansontrinh (approved):
Thank you very much for this.

PR #83, opened 2024-09-26
Rewrite BepReader to better handle concurrent modification between processes

brian-mcnamara:
With the current implementation, it is possible (and I have run into it) that the build is in the process of writing a message to the file while the reader is trying to parse the message. This results in a InvalidProtocolBufferException being thrown and while we can try to capture this, rollback the input stream and try again, I decided this can be done best with coroutines 🤓


server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BepReader.kt
```diff
@@ -0,0 +1,59 @@
...
+          FileInputStream(eventFile),
+          BuildEventStreamProtos.BuildEvent.parser(),
+        )
+      var event: BuildEventStreamProtos.BuildEvent? = null
+      do {
+        event = reader.nextMessage()
+        if (event != null) {
+          bepServer.handleBuildEventStreamProtosEvent(event)
+          setServerPid(event)
+        }
```
brian-mcnamara, line 38: Debating whether to yield here. Leaning towards not too useful

server/server/src/test/kotlin/org/jetbrains/bsp/bazel/server/bsp/utils/DelimitedMessageReaderTest.kt
```diff
@@ -0,0 +1,157 @@
...
+          ByteArrayInputStream(partial),
+          BuildEventStreamProtos.BuildEvent.parser(),
+        )
+
+      try {
+        reader.nextMessage()
+        fail("Exception expected when a partial message was sent")
+      } catch (_: InvalidProtocolBufferException) {
+        // expected
+      }
```
agluszak, line 90:
```suggestion
      assertThrows(InvalidProtocolBufferException::class.java) { reader.nextMessage() }
```

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/utils/BUILD
```diff
@@ -3,13 +3,16 @@ load("@rules_kotlin//kotlin:jvm.bzl", "kt_jvm_library")
...
+        "DelimitedMessageReader.kt",
         "InternalAspectsResolver.kt",
     ],
     visibility = ["//server/server:__subpackages__"],
     deps = [
         "//server/bazelrunner",
         "//server/commons",
         "//server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/info",
+        "@maven//:io_grpc_grpc_protobuf",
+        "@maven//:org_jetbrains_kotlinx_kotlinx_coroutines_core",
```
agluszak, line 15:
```suggestion
        "@server_maven//:org_jetbrains_kotlinx_kotlinx_coroutines_core",
```

server coroutine libraries are separate because we have problems in the client (because IJ SDK brings its own custom coroutine lib)

agluszak (commented):
Thanks, that's a great rewrite <3

server/server/src/test/kotlin/org/jetbrains/bsp/bazel/server/bsp/utils/DelimitedMessageReaderTest.kt
brian-mcnamara: I originally tried that, but nextMessage suspends, and I am blanking on how to pass the context down into the unit

brian-mcnamara:
I was thinking this morning that it would probably be nice to do a larger refactor and make the message reader return a Flow and have the BepServer collect the messages, completely removing the need for the BepReader. Maybe something I'll play around with over the weekend if I have time

server/server/src/test/kotlin/org/jetbrains/bsp/bazel/server/bsp/utils/DelimitedMessageReaderTest.kt
agluszak: Ahh, so nevermind then :) perhaps there's a method in kotest for suspending functions, but it was a nit anyway

agluszak:
Cool! Let's merge this one in the meantime:)

agluszak (approved):

PR #84, opened 2024-09-26
[feature] Machinery to highlight unknown flags (and doc) | #BAZEL-1186

mtoader:
This adds basic machinery to resolve flag names as `Symbol` and to be able to extract documentation for them. 

It also adds an annotator highlighting unknown flags.

It has halfway done flags database (pulled from `bazel help build` output only) that will be enhanced to include flags from all commands and tags and more typing of flags such in subsequent PRs

Examples:
Doc
image

Error
image

### DIFF
--- a/plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/flags/KnownFlags.kt
+++ b/plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/flags/KnownFlags.kt
@@ -1,0 +1,9229 @@
+package org.jetbrains.bazel.languages.bazelrc.flags
+
+// temporary generated from bazel help build.
+// will be followed up with a more comprehensive variant including all the flag and extra data (tags / availability for what commands / etc)
+internal val flags =
+  object {
+// unknown line:                                                            [bazel release 7.3.1]
+// unknown line: Usage: bazel build <options> <targets>
+// unknown line:
+// unknown line: Builds the specified targets, using the options.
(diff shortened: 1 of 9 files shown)


agluszak:
Oh wow, that's really impressive! Thanks!

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/flags/KnownFlags.kt
```diff
@@ -0,0 +1,9229 @@
...
+        name = "experimental_repository_cache_hardlinks",
+        description =
+          """
+          If set, the repository cache will hardlink the file in case of a cache hit, 
+          rather than copying. This is intended to save disk space.
+          """.trimIndent(),
+      )
+
+// an integer
+// default: "0"
```
agluszak, line 44: would be great to use this info too, but that can be done in a subsequent PR :)

agluszak (approved):

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/flags/KnownFlags.kt
mtoader: ooo i plan to definitely use that :) but waiting to have it done would have made this big and late. so going bit by bit

hb-man:
Wow, THAT IS really impressive! Amazing functionality, thank you for contribution!

PR #85, opened 2024-09-26
[fix] Running tests with debug | #BAZEL-768 Done

brian-mcnamara:
Related: https://youtrack.jetbrains.com/issue/BAZEL-1114/Debug-run-configuration-adds-jvmopt
Based on https://github.com/bazelbuild/bazel/blob/ff0bc4b06b2ea814244f1f181718fe404dfb1efa/src/tools/launcher/java_launcher.h#L44
It looks like the test wrapper expects this prefix since the arguments passed to the test-setup.sh are at the end of the argument list


agluszak (approved):

agluszak:
Thanks!

mtoader:
Thanks!

PR #100, opened 2024-10-12
[fix] Deps between tests and code use `//plugin-bazel/src` | #BAZEL-1308

mtoader:
This is the suggestion made on the ticket, and it seems to work w.r.t source resolution.
 I can't tell how this will affect the overall build though (local plugin run seems to work fine).


hb-man:
Thanks for the PR. 
Though, there is a chance it won't be reviewed until next Monday - most of the team is in California on Bazelcon until saturday. and might not have time for review.

PR #101, opened 2024-10-13
[feature] Update flags database | #BAZEL-1186

mtoader:
* Switched metadata store be inside annotation (matching the original annotation from bazel source)
  * Updated the db to include flags from `build`, `clean`, `fetch`, `run`, `startup`, `sync`, `test`
  * update DocumentationProvider to show more info
  * made the parser of initial data more complete
  * discovered that some metadata is not available via cli inspection (there are undocumented flags and aliases flags)
    * added a way to express this in the code


plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/flags/Annotations.kt
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * <p>An expanded option overrides previously specified options of the same name, even if it is
+   * explicitly specified. This is the original behavior and can be surprising if the user is not
+   * aware of it, which has led to several requests to change this behavior. This was discussed in
+   * the blaze team and it was decided that it is not a strong enough case to change the behavior.
+   */
+  val expandsTo: Array<String> = [],
+  /**
+   * The old name for this option. If an option has a name "foo" and an old name "bar", --foo=baz
+   * and --bar=baz will be equivalent. If the old name is used, a warning will be printed indicating
```
xuansontrinh, line 99: I don't see the logic for showing a warning for the old name yet, is it expected?
Also, with this PR, when I use the old flag name, it shows the doc with the new name, which is kinda confusing:
image

xuansontrinh (changes requested):

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/flags/Annotations.kt
mtoader: I forgot to update the doc view to show that the current flag is an alias. Will also do that.

xuansontrinh: 2 follow-up questions: considering the case where a flag name is considered "old" in Bazel X but is "new" in Bazel X-1:
- what should we show to the user?
- do you consider incorporating Bazel version of the project when showing the doc?

mtoader: pushed a change to also show the oldName

PR #132, opened 2024-11-09
Improved reporting of BEP runs with multiple test results

mworzala:
Previously a new BSP `build/taskStart` and `build/taskFinish` event was triggered with the same build identifier for each test result within a single BEP task run. The result was only showing one of the tests in the IJ test report UI.

This was encountered using the `java_test_suite` rule from [`contrib_rules_jvm`](https://github.com/bazel-contrib/rules_jvm?tab=readme-ov-file#java_test_suite) which creates a Bazel [`test_suite`](https://bazel.build/reference/be/general#test_suite) wrapping many individual tests.

My solution was to track a test run from the beginning to end BEP events which seems solid. I moved tracking of event statuses into `BspClientTestNotifier` which feels somewhat strange, however a bunch of added state in `BspServer` also seemed strange. I will wait for feedback on this.

The TestReport now contains correct metrics for the number of tests executed, which was accidentally fixed by the above change. It tracks number of leaf tests, not including wrapping suites. This behavior was taken from what the IJ task report shows, not sure if suites should be included or not.


hb-man:
please, merge latest main into all PRs, because it includes changes that require for some of the tests on CI

agluszak (approved):

PR #135, opened 2024-11-14
Fix project open processor not opening project views

brian-mcnamara:
Looks like a regression introduced in https://github.com/JetBrains/hirschgarten/commit/e02e9e36ffb6d3cacb701645fcb3fd54f0f97bba


xuansontrinh (approved):
Thank you for catching that.
It is indeed missing.

PR #137, opened 2024-11-17
[feature] Redo flags db from bazel sources | #BAZEL-1186

mtoader:
I have a way to pull flags info from the bazel sources. I want to automate it further for my benefit but in meantime i want to update the known flag to match what is in the 7.4.1 release.


agluszak (approved):

jastice:
> I have a way to pull flags info from the bazel sources
Sounds cool, so it could be done at build time?

mtoader:
> > I have a way to pull flags info from the bazel sources
> 
> Sounds cool, so it could be done at build time?

Definitely at build time, since i need to patch bazel sources for it.

PR #166, opened 2024-12-04
[feature] Special case `watchfs` when deprecation | #BAZEL-1553

mtoader:


plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/annotation/BazelrcFlagAnnotator.kt
```diff
@@ -15,6 +16,7 @@ import org.jetbrains.bazel.languages.bazelrc.flags.OptionMetadataTag
 import org.jetbrains.bazel.languages.bazelrc.highlighting.BazelrcHighlightingColors
 import org.jetbrains.bazel.languages.bazelrc.psi.BazelrcFlag
 import org.jetbrains.bazel.languages.bazelrc.psi.BazelrcLine
+import org.jetbrains.kotlin.analysis.utils.printer.parentOfType
```
agluszak, line 19: this seems to come from some internal kotlin code, could you replace it with its implementation? https://github.com/JetBrains/kotlin/blob/322a4f1e1af1ace1858ed95b42f786e15b2b9581/analysis/analysis-internal-utils/src/org/jetbrains/kotlin/analysis/utils/printer/psiUtils.kt#L14

I'd rather not use this implicit dependency

agluszak (commented):
Just one nit. Thanks!

plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/bazelrc/annotation/BazelrcFlagAnnotator.kt
mtoader: Fixed

agluszak (approved):

PR #202, opened 2025-01-05
[feature] Labels like flags are not highlighted (tests) | #BAZEL-1614

mtoader:
Add some testing for highlighting unknown flags that shows label like flags not being highlighted


jastice (approved):
Thanks a bunch!

PR #209, opened 2025-01-15
add support for lucidsoftware/rules_scala

ryan28561:
This should make it possible to run bazel-bsp with metals on projects that rely on lucidsoftware's fork of rules_scala. The main changes are:

- Tweaked the scala_info aspect and related code to support rules_scala
- Restored the dependencySources build server protocol feature (because metals is still using it)
- Added the --norun_validations flag to the aspect run to avoid accidentally pulling in slower actions as a part of the aspect run

As a note, I have tried running metals + bazel-bsp against a few different open source scala projects that were using bazel_io_rules_scala, and I couldn't get any of them to work. I did my best to "do no harm" with these changes if people are using this successfully, but if someone can provide me with steps to get a working setup, I'd appreciate any pointers. Similar comment on tests, I found and ran tests, most passed, but I'm not sure how I could have effected the ones that failed locally, so I'm hoping it's a problem with my environment and not a mistake I made with my changes.

### DIFF
--- a/server/aspects/rules/scala/scala_info.bzl
+++ b/server/aspects/rules/scala/scala_info.bzl
@@ -1,5 +1,9 @@
 load("//aspects:utils/utils.bzl", "file_location", "is_external", "map", "update_sync_output_groups")
 
+MAINLINE_SCALA_TOOLCHAIN = "@io_bazel_rules_scala//scala:toolchain_type"
+ANNEX_SCALA_TOOLCHAIN = "@rules_scala_annex//rules/scala:toolchain_type"
+
+
 SCALA_COMPILER_NAMES = [
     "scala3-compiler",
     "scala-compiler",
@@ -48,31 +52,62 @@ def extract_scalatest_classpath(rule_attr):
         extract_from_attr("_scalatest_reporter")
     )
 
-def extract_scala_info(target, ctx, output_groups, **kwargs):
+def extract_scala_info_mainline(target, ctx, output_groups):
+    scala_info = {}
+    rule_attr = ctx.rule.attr
+    common_scalac_opts = ctx.toolchains[MAINLINE_SCALA_TOOLCHAIN].scalacopts
+    if hasattr(rule_attr, "_scalac"):
+        scalac = rule_attr._scalac
+        compiler_classpath = find_scalac_classpath(scalac.default_runfiles.files.to_list())
+        if compiler_classpath:
+            scala_info["compiler_classpath"] = map(file_location, compiler_classpath)
+            if is_external(scalac):
+                update_sync_output_groups(output_groups, "bsp-sync-artifact", depset(compiler_classpath))
+    scala_info["scalac_opts"] = common_scalac_options + getattr(ctx.rule.attr, "scalacopts", [])
+    scala_info["scalatest_classpath"] = extract_scalatest_classpath(ctx.rule.attr)
+    return scala_info
+    
+def extract_scala_info_annex(target, ctx, output_groups):
+    scala_info = {}
+    rule_attr = ctx.rule.attr
+    scala_configuration = ctx.toolchains[SCALA_TOOLCHAIN].scala_configuration
+    common_scalac_options = scala_configuration.global_scalacopts
+
+    classpath_files = []
+    for target in scala_configuration.compiler_classpath:
+        for file in target[JavaInfo].runtime_output_jars:
+            classpath_files.append(file)
+    compiler_classpath = find_scalac_classpath(classpath_files)
+
+    if len(compiler_classpath) > 0:
+        scala_info["compiler_classpath"] = map(file_location, compiler_classpath)
+
+        if any([is_external(target) for target in scala_configuration.compiler_classpath]):
+            update_sync_output_groups(
+                output_groups,
+                "external-deps-resolve",
+                depset(compiler_classpath)
+            )
+
+    scala_info["scalac_opts"] = common_scalac_options + getattr(ctx.rule.attr, "scalacopts", [])
+    scala_info["scalatest_classpath"] = extract_scalatest_classpath(ctx.rule.attr)
+    return scala_info
+
+def extract_scala_info(target, ctx, output_groups):
     kind = ctx.rule.kind
     if not kind.startswith("scala_") and not kind.startswith("thrift_"):
         return None, None
-
-    SCALA_TOOLCHAIN = "@io_bazel_rules_scala//scala:toolchain_type"
-
-    scala_info = {}
 
     rule_attr = ctx.rule.attr
 
     # check of _scala_toolchain is necessary, because SCALA_TOOLCHAIN will always be present
     if hasattr(rule_attr, "_scala_toolchain"):
-        common_scalac_opts = ctx.toolchains[SCALA_TOOLCHAIN].scalacopts
-        if hasattr(rule_attr, "_scalac"):
-            scalac = rule_attr._scalac
-            compiler_classpath = find_scalac_classpath(scalac.default_runfiles.files.to_list())
-            if compiler_classpath:
-                scala_info["compiler_classpath"] = map(file_location, compiler_classpath)
-                if is_external(scalac):
-                    update_sync_output_groups(output_groups, "bsp-sync-artifact", depset(compiler_classpath))
+        scala_info = extract_scala_info_mainline(target, ctx, output_groups)
+    elif ANNEX_SCALA_TOOLCHAIN in ctx.toolchains:
+        scala_info = extract_scala_info_annex(target, ctx, output_groups)
     else:
-        common_scalac_opts = []
-    scala_info["scalac_opts"] = common_scalac_opts + getattr(ctx.rule.attr, "scalacopts", [])
-
-    scala_info["scalatest_classpath"] = extract_scalatest_classpath(rule_attr)
+        scala_info = {}
+        scala_info["scalac_opts"] =  getattr(ctx.rule.attr, "scalacopts", [])
+        scala_info["scalatest_classpath"] = extract_scalatest_classpath(rule_attr)
 
     return dict(scala_target_info = struct(**scala_info)), None
--- a/server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspAspectsManager.kt
+++ b/server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspAspectsManager.kt
@@ -73,7 +73,7 @@ class BazelBspAspectsManager(
   fun generateAspectsFromTemplates(
     ruleLanguages: List<RuleLanguage>,
     workspaceContext: WorkspaceContext,
-    toolchains: Map<RuleLanguage, Label?>,
+    toolchains: Map<RuleLanguage, List<Label>>,
     bazelRelease: BazelRelease,
     repoMapping: RepoMapping,
   ) {
@@ -97,7 +97,9 @@ class BazelBspAspectsManager(
           "javaEnabled" to javaEnabled.toString(),
           "pythonEnabled" to pythonEnabled.toString(),
           "bazel8OrAbove" to bazel8OrAbove.toString(),
-          "toolchainType" to ruleLanguage?.let { rl -> toolchains[rl]?.toString()?.let { "\"" + it + "\"" } },
+          // This will only return the first toolchain in the list.
+          // At the moment, there is only one place this is used, and it wouldn't make sense for there to be more than one anyway though.
+          "toolchainType" to ruleLanguage?.let { rl -> toolchains[rl]?.firstOrNull()?.toString().let { "\"" + it + "\"" } },
         )
       templateWriter.writeToFile(templateFilePath, outputFile, variableMap)
     }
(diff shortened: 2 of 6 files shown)


agluszak:
Hi! Thank you very much for your contribution. Please run `bazel run format` because CI is failing because of that.

Regarding upstream rules_scala support: our codebase has suffered from a bit of bitrot in this area to be honest, because we mostly focused on java + kotlin. But we absolutely want to restore full scala support in the future.

agluszak (approved):

agluszak:
LGTM, but perhaps @abrams27 wants to have a second look

agluszak:
Do you know in which Bazel version this flag was introduced exactly?  Google plugin seems to use [--noexperimental_run_validations](https://github.com/bazelbuild/intellij/blob/8984aa2ae7384ab02c9fd1a1091a2bf96427ec3e/base/src/com/google/idea/blaze/base/command/BlazeFlags.java#L61)

server/aspects/rules/scala/scala_info.bzl
```diff
@@ -48,31 +52,62 @@ def extract_scalatest_classpath(rule_attr):
...
+    
+def extract_scala_info_annex(target, ctx, output_groups):
+    scala_info = {}
+    rule_attr = ctx.rule.attr
+    scala_configuration = ctx.toolchains[SCALA_TOOLCHAIN].scala_configuration
+    common_scalac_options = scala_configuration.global_scalacopts
+
+    classpath_files = []
+    for target in scala_configuration.compiler_classpath:
+        for file in target[JavaInfo].runtime_output_jars:
```
abrams27, line 78: wont it fail for bazel 8 repos? you need to explicitly import the provider. You can use `get_java_info` from `server/aspects/rules/java/java_info.bzl.template` it has all the needed magic

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspAspectsManager.kt
```diff
@@ -156,7 +158,12 @@ class BazelBspAspectsManager(
     val allowManualTargetsSyncFlags = if (shouldSyncManualFlags) listOf(buildManualTests()) else emptyList()
     val syncFlags = workspaceContextProvider.currentWorkspaceContext().syncFlags.values
 
-    val flagsToUse = defaultFlags + allowManualTargetsSyncFlags + syncFlags
+    /*
+     * Explicitly specifying a list of output groups should prevent bazel from building the default output group, but it will still include the validations output group unless you specifically exclude it.
+     * If the validations output group depends on the results of other more costly actions, they can make this take a long time.
+     * https://bazel.build/extending/rules#validation_actions
+     */
+    val flagsToUse = defaultFlags + allowManualTargetsSyncFlags + syncFlags + listOf("--norun_validations")
```
abrams27, line 166: just a nit pick, but I'd add this flag to the `defaultFlags`

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelToolchainManager.kt
```diff
@@ -38,7 +39,7 @@ class BazelToolchainManager(private val bazelRunner: BazelRunner, private val fe
           .waitAndGetResult(cancelChecker)
           .isSuccess
       }
-    return if (androidToolchainExists) androidToolchain else NATIVE_ANDROID_TOOLCHAIN
+    return listOf(if (androidToolchainExists) androidToolchain else NATIVE_ANDROID_TOOLCHAIN).filterNotNull()
```
abrams27, line 42:
```suggestion
    return listOf(if (androidToolchainExists) androidToolchain else NATIVE_ANDROID_TOOLCHAIN)
```
as far as I can see they are not nullable

server/aspects/rules/scala/scala_info.bzl
```diff
@@ -48,31 +52,62 @@ def extract_scalatest_classpath(rule_attr):
...
+            if is_external(scalac):
+                update_sync_output_groups(output_groups, "bsp-sync-artifact", depset(compiler_classpath))
+    scala_info["scalac_opts"] = common_scalac_options + getattr(ctx.rule.attr, "scalacopts", [])
+    scala_info["scalatest_classpath"] = extract_scalatest_classpath(ctx.rule.attr)
+    return scala_info
+    
+def extract_scala_info_annex(target, ctx, output_groups):
+    scala_info = {}
+    rule_attr = ctx.rule.attr
+    scala_configuration = ctx.toolchains[SCALA_TOOLCHAIN].scala_configuration
```
abrams27, line 73: `SCALA_TOOLCHAIN` doesnt exist, sync is failing for the upstream scala rules

server/aspects/rules/scala/scala_info.bzl
```diff
@@ -48,31 +52,62 @@ def extract_scalatest_classpath(rule_attr):
...
+            scala_info["compiler_classpath"] = map(file_location, compiler_classpath)
+            if is_external(scalac):
+                update_sync_output_groups(output_groups, "bsp-sync-artifact", depset(compiler_classpath))
+    scala_info["scalac_opts"] = common_scalac_options + getattr(ctx.rule.attr, "scalacopts", [])
+    scala_info["scalatest_classpath"] = extract_scalatest_classpath(ctx.rule.attr)
+    return scala_info
+    
+def extract_scala_info_annex(target, ctx, output_groups):
+    scala_info = {}
+    rule_attr = ctx.rule.attr
```
abrams27, line 72: unused

server/aspects/rules/scala/scala_info.bzl
```diff
@@ -48,31 +52,62 @@ def extract_scalatest_classpath(rule_attr):
         extract_from_attr("_scalatest_reporter")
     )
 
-def extract_scala_info(target, ctx, output_groups, **kwargs):
+def extract_scala_info_mainline(target, ctx, output_groups):
+    scala_info = {}
+    rule_attr = ctx.rule.attr
+    common_scalac_opts = ctx.toolchains[MAINLINE_SCALA_TOOLCHAIN].scalacopts
```
abrams27, line 58:
```suggestion
    common_scalac_options = ctx.toolchains[MAINLINE_SCALA_TOOLCHAIN].scalacopts
```

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelToolchainManager.kt
```diff
@@ -3,28 +3,29 @@ package org.jetbrains.bsp.bazel.server.bsp.managers
...
 
 class BazelToolchainManager(private val bazelRunner: BazelRunner, private val featureFlags: FeatureFlags) {
-  fun getToolchain(ruleLanguage: RuleLanguage, cancelChecker: CancelChecker): Label? =
+  fun getToolchain(ruleLanguage: RuleLanguage, cancelChecker: CancelChecker): List<Label> =
     when (ruleLanguage.language) {
-      Language.Scala -> Label.parse("@io_bazel_rules_scala//scala:toolchain_type")
-      Language.Java -> Label.parse("@bazel_tools//tools/jdk:runtime_toolchain_type")
-      Language.Kotlin -> Label.parse("@${ruleLanguage.ruleName}//kotlin/internal:kt_toolchain_type")
-      Language.Rust -> Label.parse("@${ruleLanguage.ruleName}//rust:toolchain_type")
+      Language.Scala -> listOf(Label.parse("@io_bazel_rules_scala//scala:toolchain_type"), Label.parse("@rules_scala_annex//rules/scala:toolchain_type"))
```
abrams27, line 12: we need to conditionally add this toolchain only if it's available, currently it breaks the upstream rules import (repo doesnt exist)

abrams27 (changes requested):
thanks a lot for the contribution! I found some issues and they probably explain why some tests are failing

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/languages/java/JavaLanguagePlugin.kt
```diff
@@ -53,7 +56,23 @@ class JavaLanguagePlugin(
   private fun getJdk(): Jdk = jdk ?: throw RuntimeException("Failed to resolve JDK for project")
 
   override fun dependencySources(targetInfo: TargetInfo, dependencyGraph: DependencyGraph): Set<URI> =
-    emptySet() // Provided via workspace/libraries
+    targetInfo.getJvmTargetInfoOrNull()?.run {
```
abrams27, line 59: and I'd do this calculation conditional as well - it's quite heavy calculation executed for each target, what can add quite some overhead for us (we use a separate mechanism for it, because this one is too slow for us).

By "conditional" everywhere I meant that it's possible to check what rules are imported (org/jetbrains/bsp/bazel/server/sync/ProjectResolver.kt:69), it's a list of all rules in the workspace, and then you can use this list to check if the fork rules are loaded

ryan28561:
Thanks for your feedback! I'm a little slow, but I'll get to work on these!

ryan28561:
> Do you know in which Bazel version this flag was introduced exactly? Google plugin seems to use [--noexperimental_run_validations](https://github.com/bazelbuild/intellij/blob/8984aa2ae7384ab02c9fd1a1091a2bf96427ec3e/base/src/com/google/idea/blaze/base/command/BlazeFlags.java#L61)

It looks to me like it was renamed in 5.0. https://blog.bazel.build/2022/01/19/bazel-5.0.html

abrams27:
> It looks to me like it was renamed in 5.0

nice, so we dont have to worry about it, bazel 5 is deprecated

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/languages/java/JavaLanguagePlugin.kt
ryan28561: Ok, so on this one I have a question. As far as I can tell, the `workspace/libraries` command isn't a part of the [build server protocol](https://build-server-protocol.github.io/docs/specification) that I can see, and more practically speaking, metals lsp doesn't seem to consume it. All that is to say, I enabled this more for metals support than for rules_scala_annex support. 

I'm happy to do what you recommended just to make my use case work, but is it possible we could make this something configurable instead? I see there are some FeatureFlags in the repo, would that be a viable place to add a toggle for something like this?

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelToolchainManager.kt
```diff
@@ -3,12 +3,17 @@ package org.jetbrains.bsp.bazel.server.bsp.managers
...
 import org.jetbrains.bsp.protocol.FeatureFlags
 
 class BazelToolchainManager(private val bazelRunner: BazelRunner, private val featureFlags: FeatureFlags) {
   fun getToolchain(ruleLanguage: RuleLanguage, cancelChecker: CancelChecker): Label? =
     when (ruleLanguage.language) {
-      Language.Scala -> Label.parse("@io_bazel_rules_scala//scala:toolchain_type")
+      Language.Scala ->
+        when (ruleLanguage.ruleName) {
+          "rules_scala_annex" -> Label.parse("@rules_scala_annex//rules/scala:toolchain_type")
+          else -> Label.parse("@${ruleLanguage.ruleName}//scala:toolchain_type")
```
abrams27, line 15: it 
```suggestion
          else -> Label.parse("@io_bazel_rules_scala//scala:toolchain_type")
```

even in bzlmod, it's still prefixed with `io_bazel`

abrams27, line 15: it 
```suggestion
          else -> Label.parse("@io_bazel_rules_scala//scala:toolchain_type")
```

even in bzlmod, it's still prefixed with `io_bazel`

abrams27 (changes requested):
and it seems like not everything is building

server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/sync/languages/java/JavaLanguagePlugin.kt
ryan28561: > override it only in our client

I'm not sure how to do that. Does that need to happen in another repo?

abrams27:
please run `bazel //:format` to make the formatter happy

jastice:
Hi @ryan28561 sorry this PR didn't make it in yet. We now have a very different situation - we have removed the BSP support completely, so it probably won't be very useful for you to merge this. If you require it for Scala support in IntelliJ however, we can make an effort to fix the conflicts and merge it after all. Please let us know.

The latest commit before BSP removal should be this one:
https://github.com/JetBrains/hirschgarten/commit/bdbf646721961fe1212e228efd84a8825da43aac

ryan28561:
Hey, thanks for the heads up @jastice . Can you give any clarification on why bsp support was removed and if there's a chance that it will be brought back in the future?

jastice:
We determined that at the moment, BSP support is only an overhead for the development of good Bazel support in IntelliJ. We would be open bringing it back if we can get more build tool vendors interested in implementing it, but in that case probably not as part of this project directly.

PR #210, opened 2025-01-22
Update python_info aspect to accommodate Python targets that only have rules_python providers

mnoah1:
As the Python rules are gradually phasing out use of the builtin PyInfo provider in favor of the ones from rules_python, there could be cases where a repository is on a Bazel version lower than 8, but the built in PyInfo is still missing from a given target.

In our repo's case, we are in the process of migrating to rules_py, which uses the rules_python providers only (https://github.com/aspect-build/rules_py/pull/311), and not the built in ones.

This change adds support for rules_python providers in the aspect for Bazel versions less than 8:
- Bazel <8 - this adds an extra check rules_python in the repo's external rules.  The load statement will only run if present, and when this is the case, the rules_python providers will be considered as a secondary option to the builtin ones.
- Bazel 8 - rules_python providers only [existing behavior]


server/aspects/rules/python/python_info.bzl.template
```diff
@@ -1,20 +1,23 @@
-#if( $pythonEnabled == "true" && $bazel8OrAbove == "true" )
+#if( $pythonEnabled == "true" )
+#if( $bazel8OrAbove == "true")
 load("@rules_python//python:defs.bzl", "PyInfo", "PyRuntimeInfo")
 #end
+load("@rules_python//python:defs.bzl", PyInfoRulesPython = "PyInfo", PyRuntimeInfoRulesPython = "PyRuntimeInfo")
```
agluszak, line 5: Are you sure this is correct? You load the same PyInfo twice if `bazel8OrAbove == "true"`

agluszak (changes requested):

server/aspects/rules/python/python_info.bzl.template
mnoah1: Loading again in this case would be ok since it's under a different alias (originally I was thinking of trying to avoid adding to much additional logic to the conditionals in the file).

However I found that this approach does not properly accommodate Bazel versions <7, which could fail if the workspace did not explicitly include rules_python.  I made some updates to handle this case and make the rest of the logic clearer.

mnoah1:
It looks like all the e2e tests pass with this change, but I'm seeing a failure in the `
[analysis] Qodana android-forked-bazel-rules-project` step with `error: found 'project contains no modules' in the build log`.  I don't think it seems related to my change but I'm not quite sure - could you help me verify?

mnoah1:
Hi @agluszak - could you give this one another look when you have a chance?  Thanks.

agluszak (approved):

mnoah1:
> could you please merge main for it to pass?
Updated - looks like the new one is passing.

> regarding [analysis] Qodana android-forked-bazel-rules-project - we are investigating, but it seems your change indeed breaks something in one of our android project e2e test
Thanks - let me know if you find any more specifics (I was able to easily run the other e2e tests locally, but seems the Qodana ones require a token and are a bit harder for me to debug myself).

gottagofaster236:
2025-02-07T12:42:51.793Z [13336] INFO - [31m[1mERROR: [0mat /root/.cache/bazel/_bazel_idea/d1aa40fbc8d41d43ca12eb8239db2224/external/bazelbsp_aspect/aspects/core.bzl:3:6: at /root/.cache/bazel/_bazel_idea/d1aa40fbc8d41d43ca12eb8239db2224/external/bazelbsp_aspect/aspects/extensions.bzl:3:6: at /root/.cache/bazel/_bazel_idea/d1aa40fbc8d41d43ca12eb8239db2224/external/bazelbsp_aspect/aspects/rules/python/python_info.bzl:4:6: Unable to find package for @@rules_python//python:defs.bzl: The repository '@@rules_python' could not be resolved: Repository '@@rules_python' is not defined.

server/aspects/rules/python/python_info.bzl.template
```diff
@@ -1,20 +1,35 @@
...
+# Bazel 8+: Use only the rules_python providers.
 load("@rules_python//python:defs.bzl", "PyInfo", "PyRuntimeInfo")
-#end
+PyInfo_Secondary = None
+PyRuntimeInfo_Secondary = None
 
-load("//aspects:utils/utils.bzl", "create_struct", "file_location", "to_file_location")
+#elseif( $bazel7OrAbove == "true" && $bazel8OrAbove == "false")
+# Bazel 7: Keep builtin providers available with their original names, and import rules_python providers as secondary.
+load("@rules_python//python:defs.bzl", PyInfo_Secondary = "PyInfo", PyRuntimeInfo_Secondary = "PyRuntimeInfo")
```
gottagofaster236, line 14: I'm guessing this is the line that is failing the test, are you sure this load statement will always succeed?

mnoah1: Thanks for checking - it does appear that there are cases where this can fail.   I'm not sure of the exact behavior, but all repos in the e2e tests section seem to have rules_python via dependencies included in DEFAULT.WORKSPACE.SUFFIX, however for android-forked-bazel-rules-project that does not seem to be the case. It seems like it will be better to just move this check to the part where external rules are discovered, to ensure that it is really there (https://github.com/JetBrains/hirschgarten/blob/f4f118c49014e1aecfd53110aefaf69d5dac67ac/server/server/src/main/kotlin/org/jetbrains/bsp/bazel/server/bsp/managers/BazelBspAspectsManager.kt#L55).

mnoah1:
This is now updated to better handle possible scenarios, and all checks are now passing.  This is ready for another look when you have a chance.

mnoah1:
@agluszak @jastice @gottagofaster236 just wanted to follow up on this.

xuansontrinh:
@mnoah1 the PR LGTM.
Could you please resolve the conflict with the readme file and then we can merge it?

PR #223, opened 2025-02-23
Restore `.bsp/bazelbsp.json` on install

harpocrates:
This was recently removed in 0b6f5fdf001880100. Adding it back just because this is a pretty important part of how a tool consuming a BSP implementation can generically start the server. Other BSP implementations (like SBT) also generate the JSON file.

### DIFF
--- a/server/executioncontext/installationcontext/BUILD
+++ b/server/executioncontext/installationcontext/BUILD
@@ -1,0 +1,9 @@
+load("@rules_kotlin//kotlin:jvm.bzl", "kt_jvm_library")
+
+kt_jvm_library(
+    name = "installationcontext",
+    visibility = ["//visibility:public"],
+    exports = [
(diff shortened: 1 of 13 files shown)


harpocrates:
Please let me know if I've misunderstood something about how the `bazel-bsp` BSP implementation is supposed to be used - the commit I linked seems to be hinting at big changes. The main thrust of it seemed to be about removing the IJ `bsp` plugin, but the `.bsp/bazelbsp.json` file is useful for more than just that plugin.

agluszak:
Hi. We're completely removing BSP support from Hirschgarten. As all other major implementations of BSP are for Scala, the support of the generic BSP will most likely return to https://github.com/jetBrains/intellij-scala

jastice:
Adding to that, bazel-bsp server will be discontinued in this repo. If you would like to take over maintenance, we can talk about how to organize this.

PR #224, opened 2025-02-23
[feature] Add quick fixes to old and no-op flags | #BAZEL-1704 Done

mtoader:
Add some quick fixes to deprecated and old flag names. With tests 

This still uses the 7.4.1 flags db so it might be incorrect on newer bazel. I need to get access to the bazel version in order to use a accurate db.


agluszak (approved):
Lgtm

PR #229, opened 2025-03-10
[feature] Add flags completion | #BAZEL-1186

mtoader:
* update the flags db to match bazel 8.4.1 (still need a solution that can let me switch the db based on actual project version: see BAZEL-1703)
  * add list of commands a flag is applicable and update annotator completion to use it


agluszak (approved):

agluszak:
Thanks!

mtoader:
> * BAZEL-1703

How can I get some help with BAZEL-1703? It would help to tailor the flag db to the actual bazel version of the project.

PR #230, opened 2025-03-11
Kl/create bazel plugin zip

kateliu20:


MODULE.bazel
```diff
@@ -192,6 +194,9 @@ use_repo(
     "build_bazel_bazel_8_0_0",
 )
 
+apple_cc_configure = use_extension("@apple_support//crosstoll:setup.bzl", "apple_cc_configure_extension")
+use_repo(apple_cc_configure, "local_config_apple_cc")
+
```
gottagofaster236, line 199: I think the plugin builds fine on macOS, or did it not work for you without this change?

server/resources/aspects/rules/android/android_info.bzl.template
```diff
@@ -106,17 +107,17 @@ def extract_android_aar_import_info(target, ctx, dep_targets, **kwargs):
...
+    if AndroidIdeInfo not in target:
         return None, None
-    manifest = file_location(target[AndroidManifestInfo].manifest)
+    manifest = file_location(target[AndroidIdeInfo].manifest)
 
     resource_folder = None
     r_txt = None
     if AndroidResourcesInfo in target:
-        direct_android_resources = target[AndroidResourcesInfo].direct_android_resources.to_list()
+        direct_android_resources = target[AndroidResourcesInfo].direct_resources_nodes.to_list()
```
gottagofaster236, line 117: Could you do an if-condition, e.g. `if hasattr(target[AndroidResourcesInfo], "direct_resources_nodes"):` <new code> `else:` <old code>  so that both old and new versions of `rules_android` are supported?

server/resources/aspects/rules/android/android_info.bzl.template
```diff
@@ -106,17 +107,17 @@ def extract_android_aar_import_info(target, ctx, dep_targets, **kwargs):
...
         if direct_android_resources:
             direct_android_resource = direct_android_resources[0]
-            resource_files = direct_android_resource.resources
+            resource_files = direct_android_resource.resource_files.to_list()
```
gottagofaster236, line 120: Same here

gottagofaster236 (changes requested):
Thx for making the PR!!

server/resources/aspects/rules/android/android_info.bzl.template
```diff
@@ -1,4 +1,5 @@
 load("//aspects:utils/utils.bzl", "create_file_location", "create_struct", "file_location")
+load("@rules_android//providers:providers.bzl", "AndroidIdeInfo", "AndroidManifestInfo", AndroidResourcesInfo = "StarlarkAndroidResourcesInfo")
```
gottagofaster236, line 2: This one is a bit more difficult to fix :( 

1. So before the import was `@rules_android//rules:providers.bzl`, and then 6 months ago they [changed](https://github.com/bazelbuild/rules_android/commit/0f1caa9d42e5d128b894509a59aa21b3a797b0e3) it to `@rules_android//providers:providers.bzl`. The way to figure out which one is needed would be by just checking it in the file system. You can get the location of where `rules_android` are unpacked from `repoMapping`:

   image
   With that you can just check which one of two paths actually exist on disk (relative to the root of `rules_android` repository).
   Then the correct path (`providers:providers.bzl` or `rules:providers.bzl`) could be passed via our templating mechanism (see `BazelBspAspectsManager.kt`).

2. It's actually not always `rules_android`, it can also be `build_bazel_rules_android`!
   image
   Instead, you can use `@${rulesetName}` which is then expanded when the aspect is written on disk. Here's an example:
   image
   
3. We also support(ed) native Android rules (for Bazel 7 or for Bazel 8 with autoloads). I think you could use this `$rulesetName != ""` check like above, i.e. 
   ```
   #if( $rulesetName != "" )
   load("@${rulesetName}//${providerPath}", "AndroidIdeInfo", "AndroidManifestInfo", AndroidResourcesInfo = "StarlarkAndroidResourcesInfo")
   #end

PR #245, opened 2025-04-17
[feature] Recognize keyword-only boundary marker in parameter lists

fmeum:
This avoids showing a syntax error for `def f(x, *, y)`, but still (correctly) shows one for `def f(x, *)`.


gottagofaster236:
Could you please add a test that checks that the case of `def f(x, *)` is still an error?

fmeum:
Added

gottagofaster236 (approved):

gottagofaster236:
Thx a lot for your PRs!!

PR #253, opened 2025-04-22
[fix] Take scopes into account when resolving Starlark references

fmeum:


plugin-bazel/src/main/kotlin/org/jetbrains/bazel/languages/starlark/psi/statements/StarlarkStatementContainer.kt
```diff
@@ -0,0 +1,5 @@
+package org.jetbrains.bazel.languages.starlark.psi.statements
+
+interface StarlarkStatementContainer {
+  fun getStatementLists(): List<StarlarkStatementList>
```
agluszak, line 4: why are there multiple lists?

fmeum: This is in draft mode and not quite ready for review yet, but thanks for taking a look.

An `if` statement has multiple child `StarlarkStatementList`s, one for each branch.

hb-man:
@fmeum it fails during crossbuild for older IDEA version
try `bazel build //... --define=ij_product=intellij-2024.3` and you should be able to reproduce the error.

fmeum:
@hb-man Thanks for the pointer, should be fixed now.

xuansontrinh:
@fmeum could you please merge the latest `main` to this PR?
I will merge it after.

fmeum:
> some Starlark unittests are still failing.
> Could you please check?

Should be fixed now, commits merged in the meantime required changes

xuansontrinh:
merging the PR.
Thank you for your contribution!

PR #254, opened 2025-04-28
[fix] Don't use `--inject_repository` with `--noenable_bzlmod`

fmeum:
`--inject_repository` is a no-op with `--noenable_bzlmod` and `--override_repository` should be used instead.


hb-man:
(the one test failing is currently a problem on our side, you can ignore it)

xuansontrinh (approved):

PR #276, opened 2025-06-01
Don't mark non-top-level elements with `_` prefix as unused

fmeum:
Using `_` as a prefix of an identifier that isn't top-level indicates that it is expected to be unused.


hb-man:
thank you, merged into main in 74869773fb4d0cda806593fc1a4559ba695810d7

PR #346, opened 2025-11-17
Fix StackOverflowError during "Apply changes to the internal targets model"

evis:
Problem: `calculateOutputJarsLibraries()` was creating synthetic dependencies from targets to output_jars libraries that already exist as real targets. For example, with BUILD file:

```starlark
proto_library(
    name = "proto",
    srcs = glob(["proto/**/*.proto"]),
    strip_import_prefix = "proto",
)

java_proto_library(
    name = "proto_output_jars",
    deps = [":proto"],
)
```

The plugin would create a synthetic dependency proto -> proto_output_jars, combined with the real dependency proto_output_jars -> proto, creating cycles: proto -> proto_output_jars -> proto -> proto_output_jars -> ...

This resulted in infinite recursion and `StackOverflowError` in `calculateTransitivelyExecutableTargets` during the "Apply changes to the internal targets model" sync phase, causing sync to hang.

Stack trace excerpt:
```
at org.jetbrains.bazel.target.TargetUtils.calculateTransitivelyExecutableTargets(TargetUtils.kt:206)
at org.jetbrains.bazel.target.TargetUtils.access$calculateTransitivelyExecutableTargets(TargetUtils.kt:60)
at org.jetbrains.bazel.target.TargetUtils$calculateTransitivelyExecutableTargets$1$1.invoke(TargetUtils.kt:216)
at org.jetbrains.bazel.target.TargetUtils$calculateTransitivelyExecutableTargets$1$1.invoke(TargetUtils.kt:215)
[infinite recursion continues...]
```

Solution: Don't create dependency from target to synthetic output_jars library if real target already exists.


plugin-bazel/src/main/kotlin/org/jetbrains/bazel/sync/workspace/mapper/normal/AspectBazelProjectMapper.kt
```diff
@@ -254,14 +254,25 @@ class AspectBazelProjectMapper(
...
       .filter { shouldCreateOutputJarsLibrary(it) }
       .mapNotNull { target ->
-        createLibrary(Label.parse(target.id + "_output_jars"), target, onlyOutputJars = true, isInternalTarget = true)?.let { library ->
-          target.label() to listOf(library)
+        val outputJarsLabel = Label.parse(target.id + "_output_jars")
+
+        // Don't create dependency to synthetic output_jars library if real target already exists
+        // This prevents cyclic dependencies like proto -> proto_output_jars -> proto -> proto_output_jars
+        if (outputJarsLabel in targetsToImportSet) {
+          null
```
evis: This line completely prevents creating the synthetic dependency `proto -> proto_output_jars` when a real `proto_output_jars` target exists. While this fixes the `StackOverflowError`, it causes IDE issues - proto imports now show as unresolved because they can't see the generated Java classes.

The dilemma is:
- **With dependency:** `proto -> proto_output_jars` + `proto_output_jars -> proto` = cycle → StackOverflowError in `calculateTransitivelyExecutableTargets`
- **Without dependency:** No cycle, but IDE loses the connection between proto files and generated Java classes

What would be the best approach here to fix it?

1. Create the dependency but mark it as "non-transitive" for executable target calculation?
2. Fix `calculateTransitivelyExecutableTargets` to handle cycles with recursion protection?
3. Something else?

jastice:
Since this fix seems to target your specific rules we can't merge it, and it would be necessary to fix on repo side.

PR #348, opened 2025-11-26
BAZEL-2666 Add support for scala compiler plugins

kusaeva:
This PR adds support for Scalac compiler plugins (e.g., better-monadic-for) with automatic path resolution.

### DIFF
--- a/server/resources/aspects/rules/scala/scala_info.bzl.template
+++ b/server/resources/aspects/rules/scala/scala_info.bzl.template
@@ -48,6 +48,16 @@ def extract_scalatest_classpath(rule_attr):
         extract_from_attr("_scalatest_reporter")
     )
 
+def get_plugins(rule_attr):
+    result = []
+    plugins = getattr(rule_attr, "plugins", [])
+    for plugin in plugins:
+        if JavaInfo in plugin:
+            jars = getattr(plugin[JavaInfo], "full_compile_jars", depset()).to_list()
+            for f in map(file_location, jars):
+                result.append("-Xplugin:%s/%s" % (f.root_execution_path_fragment, f.relative_path))
+    return result
+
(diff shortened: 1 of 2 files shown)


aehlig (approved):
LGTM. Thanks.

server/resources/aspects/rules/scala/scala_info.bzl.template
```diff
@@ -48,6 +48,16 @@ def extract_scalatest_classpath(rule_attr):
         extract_from_attr("_scalatest_reporter")
     )
 
+def get_plugins(rule_attr):
+    result = []
+    plugins = getattr(rule_attr, "plugins", [])
+    for plugin in plugins:
+        if JavaInfo in plugin:
```
aehlig, line 55: I missed this, when I was first reviewing this change: `JavaInfo` is no longer a built-in provider, hence this is an undefined symbol when using `bazel` 9. As far as I can see, `rules_scala` does not reexport `JavaInfo` either, so there is no easy way to import it from there. Hence this computation cannot be done in this file which is supposed to `load` only from `rules_scala`.

However, the information about `full_compile_jars` is available in the `java_provider` message for each target, so this computation can (and must) be done in the plugin where information collected from providers of different rules can be combined.

aehlig (changes requested):
Sorry again for taking so long and not spotting this problem earlier.

server/resources/aspects/rules/scala/scala_info.bzl.template
jastice: Since this fix is non-trival, we unfortunately can't merge the change at this moment. Therefore I'm closing the PR, but the issue will remain open, and we'll welcome a new PR, which we'll also review more quickly :)

PR #368, opened 2025-12-01
Fix transitive proto_library dependencies for scala_proto_library

evis:
Resolves issue where imports from transitive proto_library dependencies were not resolved in IntelliJ.

The plugin now correctly adds scalapb JAR files from transitive proto_library dependencies to the module classpath, enabling proper import resolution for scala_proto_library targets with nested proto dependencies.

Fixes [BAZEL-2695](https://youtrack.jetbrains.com/issue/BAZEL-2695)

## Impact

With this PR, [this transitive proto import](https://github.com/evis/transitive-proto-reproduce/blob/main/app/Main.scala#L3) is now resolved correctly and no longer shows as red in IntelliJ.

**Scale of the problem:**
- In just one service (~220k lines of code), this PR fixes **679 red import cases**
- Across our entire repository (~4 million lines of code), the number of affected imports is much higher

This is a significant developer experience improvement for our team. In our Scala code, it's very common to use classes from transitive proto_library dependencies, and currently all such imports show as unresolved.

This has been a persistent pain point for developers working with Scala and protobuf, as it makes code navigation much more difficult when imports appear broken even though they compile correctly with Bazel.

### DIFF
--- a/plugin-bazel/src/main/kotlin/org/jetbrains/bazel/sync/workspace/mapper/normal/AspectBazelProjectMapper.kt
+++ b/plugin-bazel/src/main/kotlin/org/jetbrains/bazel/sync/workspace/mapper/normal/AspectBazelProjectMapper.kt
@@ -129,11 +129,11 @@ class AspectBazelProjectMapper(
       }
     val interfacesAndBinariesFromTargetsToImport =
       measure("Collect interfaces and classes from targets to import") {
-        collectInterfacesAndClasses(targetsToImport)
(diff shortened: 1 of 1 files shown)


gottagofaster236:
@noxvost I think this will conflict with your LanguagePlugin refactoring

noxvost:
Thank you for the solution draft. Since we are in the middle of big refactoring of the importing engine, this PR cannot be accepted as is. The mentioned issue is planned for IJ 2026.2 release cycle, and we will incorporate your ideas to the fix.

PR #380, opened 2026-03-18
[python]: indexing support for `rules_pycross` dependencies and other generated Python sources

tobyh-canva:
This PR works towards addressing the limitations of the Bazel plugin with generated Python sources, as demonstrated in https://github.com/tobyh-canva/hirschgarten_python_test.

## Testing
Since IDE sync and tests don't work on this open-source repo, I've only tested this against our own internal monorepo at Canva, by building the plugin with my change from the `253` branch. I haven't been able to do extensive testing on other monorepos.


noxvost:
PR is merged

PR #385, opened 2026-03-30
fix(bsp): Bazel 7.6.1 + WORKSPACE: batch-retry show_repo on WORKSPACE bad-repo errors

greggdonovan:
## Summary
- When WORKSPACE is enabled, `bazel mod show_repo` validates all repo arguments upfront and fails on the first unknown repo without producing any output. Previously this triggered N individual `bazel mod show_repo` calls (one per repo, running serially). In our monorepo this took ~40 minutes, now it takes 3 minutes.
- Now we parse the error to identify the bad repo, remove it from the batch, and retry — reducing calls from O(N) to O(B+1) where B is the number of bad repos.
- Falls back to the existing per-repo resolution for unrecognized errors.

## Environment

- We are running Bazel 7.6.1 with `WORKSPACE` and `MODULE.bazel`. We're doing our best to migrate to Bazel 8.x, but this is slowing us down enough that we wanted to fix it before our migration finishes.

## Workaround for
https://github.com/bazelbuild/bazel/issues/28601

### DIFF
--- a/intellij.bazel.connector/src/org/jetbrains/bazel/bazelrunner/ModuleResolver.kt
+++ b/intellij.bazel.connector/src/org/jetbrains/bazel/bazelrunner/ModuleResolver.kt
@@ -1,6 +1,7 @@
 package org.jetbrains.bazel.bazelrunner
 
 import org.jetbrains.annotations.ApiStatus
+import org.jetbrains.annotations.VisibleForTesting
 import org.jetbrains.bazel.commons.BazelInfo
 import org.jetbrains.bazel.commons.gson.bazelGson
 import org.jetbrains.bazel.workspacecontext.WorkspaceContext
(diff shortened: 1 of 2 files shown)


intellij.bazel.connector/src/org/jetbrains/bazel/bazelrunner/ModuleResolver.kt
```diff
@@ -162,19 +163,42 @@ internal class ModuleResolver(
...
         return mapOf(moduleNames[0] to null)
       }
-      return moduleNames.map { resolveModules(listOf(it), bazelInfo) }.reduceOrNull { acc, result -> acc + result } ?: emptyMap()
+      return resolveModulesIndividually(moduleNames, bazelInfo)
     }
 
     return moduleOutputParser.parseShowRepoResults(processResult, json_output)
   }
 
+  private suspend fun resolveModulesIndividually(moduleNames: List<String>, bazelInfo: BazelInfo): Map<String, ShowRepoResult?> =
```
aehlig, line 184: Why move this to a separate function instead of leaving inline?

intellij.bazel.connector/src/org/jetbrains/bazel/bazelrunner/ModuleResolver.kt
```diff
@@ -162,19 +163,42 @@ internal class ModuleResolver(
...
-      // As we cannot reliably determine which repositories `bazel mod show_repo` is willing to resolve,
-      // we have to accept failure. However, to get the maximal amount of information possible, we should
-      // ask for each repository individually.
+      // Parse the error to identify the bad repo and retry the batch without it,
+      // falling back to per-repo resolution for unrecognized errors.
+      val badRepo = extractBadRepoFromError(processResult)
+      if (badRepo != null && moduleNames.size > 1 && badRepo in moduleNames) {
+        val remaining = moduleNames.filter { it != badRepo }
+        val retryResults = resolveModules(remaining, bazelInfo)
+        return retryResults + mapOf(badRepo to null)
```
aehlig, line 172: `resolveModules` now returns a different type so this addition does not work; please rebase you change on top of current branch 261.

pluginTests/test/org/jetbrains/bazel/bazelrunner/ModuleResolverTest.kt
```diff
@@ -14,6 +24,79 @@ class ModuleResolverTest {
...
+      importDepth = 0,
+      enabledRules = emptyList(),
+      ideJavaHomeOverride = null,
+      shardSync = false,
+      targetShardSize = 0,
+      shardingApproach = null,
+      importRunConfigurations = emptyList(),
+      gazelleTarget = null,
+      indexAllFilesInDirectories = false,
+      pythonCodeGeneratorRuleNames = emptyList(),
```
aehlig, line 47: That parameter has been removed. Please rebase your change and make sure it builds on top of the current branch `261`.

pluginTests/test/org/jetbrains/bazel/bazelrunner/ModuleResolverTest.kt
```diff
@@ -14,6 +24,79 @@ class ModuleResolverTest {
     }
 
   val moduleOutputParser = ModuleOutputParser()
+  private val tempDir = Path.of(System.getProperty("java.io.tmpdir"))
```
aehlig, line 27: Instead of using `java.io.tmpdir`, please use relative paths when mocking `BazelInfo`, just as in the other tests.

aehlig (commented):
Generally I like the idea of just singling out the broken repository and repeat the batch questioning. However, the change needs to be rebased to work with the current version of `261`.

aehlig (changes requested):

nadavwe:
Hey, we have a similar problem only that we have no modules at all, only legacy repos. and we have hundreds of them. (we just started using bzlmod and we're transitioning).
so that fix would leave us with O(n).
@aehlig I thought about partitioning the repos into bzlmod and legacy - sadly I couldn't find a reliable bazel command to use for that. 
how would you feel about partitioning by the name containing '+' (for bazel >=8) or '~' (for bazel <8)?

aehlig:
> Hey, we have a similar problem only that we have no modules at all, only legacy repos. and we have hundreds of them. (we just started using bzlmod and we're transitioning). so that fix would leave us with O(n). 

As mentioned, I'm not against using a heuristic to exclude the faulty repository; however, the change would have to be rebased (and the other review comments addressed).

> @aehlig I thought about partitioning the repos into bzlmod and legacy - sadly I couldn't find a reliable bazel command to use for that. how would you feel about partitioning by the name containing '+' (for bazel >=8) or '~' (for bazel <8)?

I don't mind a heuristics for the about 8 months that bazel 7 will still be around, as long as it behaves somewhat reasonably even if the underlying assumptions fail (which unlikely for bazel 7 anyway, as there is no active feature development there).

pluginTests/test/org/jetbrains/bazel/bazelrunner/ModuleResolverTest.kt
gottagofaster236: FYI, this test is not compilable externally because of hirschgarten repo limitations, we have to adapt the test when merging in any case

greggdonovan:
We have finished our migration to Bazel 8 and removed our WORKSPACE usage faster than expected. If someone else is running into this issue and wants to move the PR forward, that would be great. Otherwise, feel free to close it. Thanks for taking a look!

PR #388, opened 2026-04-28
BAZEL-3128: Bazel 7 native protobuf support for sync aspects

dkBrazz:
For Bazel versions before 8, protobuf is treated as a bundled/native language so proto_library targets are imported even without rules_proto or protobuf declared in MODULE.bazel. Bazel 8+ continues using the existing external ruleset discovery path.

Also adds focused tests for protobuf’s pre-8 bundled behavior and preserves Bazel 8 autoload handling for Java/Python symbol hints.


aehlig (approved):
LGTM. Thanks.

aehlig:
Merged to `main` and cherry-picked to 261. Thanks for you contribution.

PR #389, opened 2026-05-06
BAZEL-3149: upgrade to toplevel only if remote_download_outputs=minimal

christianladam:
For Bazel builds that require all, the toplevel setting can break things. By only overriding remote_download_outputs when it was minimal we avoid causing problems in those specific cases.

The execute service specifically only pulls the test xml, log, and lcov outputs, so a simple remote_download_regex works in that case and is less changes.


aehlig (approved):
LGTM. Thanks.

aehlig:
Merged.

PR #390, opened 2026-05-06
BAZEL-3151: add sibling data to avoid class name collisions due to BAZEL-3117

christianladam:
If a build file isn't fine-grained per package then the current jdeps logic can result in a collision if two or more packages under a BUILD file have a class with the same name.

foo/src/main/java/BUILD (globs **/*.java, lists files individually, etc)
foo/src/main/java/com/foo/AClass.java
foo/src/main/java/com/foo/bar/AClass.java

### DIFF
--- a/intellij.bazel.connector/resources/aspects/rules/java/java_info.bzl.template
+++ b/intellij.bazel.connector/resources/aspects/rules/java/java_info.bzl.template
@@ -64,7 +64,7 @@ def get_jdeps(ctx, target):
     jdeps = get_raw_jdeps(target)
     materialized_jdeps = []
     for raw_jdeps_file in jdeps:
-        materialized_jdeps_file = ctx.actions.declare_file("materialized_" + raw_jdeps_file.basename)
+        materialized_jdeps_file = ctx.actions.declare_file("materialized_" + raw_jdeps_file.basename, sibling = raw_jdeps_file)
         ctx.actions.run_shell(
             inputs = [raw_jdeps_file],
             outputs = [materialized_jdeps_file],


aehlig (approved):
The change as is looks good to me; so if you mark it as ready, I will try to import it.

aehlig (commented):
I just realized that simply using a sibling is not possible, if rules are used that forward `jdeps` files. I suggest that I just write a different patch that, in the current package, creates a sufficiently unique file name and, once merged, will close this PR.

intellij.bazel.connector/resources/aspects/rules/java/java_info.bzl.template
```diff
@@ -64,7 +64,7 @@ def get_jdeps(ctx, target):
     jdeps = get_raw_jdeps(target)
     materialized_jdeps = []
     for raw_jdeps_file in jdeps:
-        materialized_jdeps_file = ctx.actions.declare_file("materialized_" + raw_jdeps_file.basename)
+        materialized_jdeps_file = ctx.actions.declare_file("materialized_" + raw_jdeps_file.basename, sibling = raw_jdeps_file)
```
aehlig, line 67: I just realized that using making the materialized `jdeps` file a sibling of the original, while natural and working with natural rule sets, causes illegal file declarations when rules are used that [merge and forward `JavaInfo` form dependencies](https://github.com/JetBrains/hirschgarten/blob/a608db2a0f5d54e690d794f26838fff91af1dc5a/rules_intellij/build_defs/intellij_plugin.bzl#L79): then the target has a `jdeps` file taken from a dependency which might be of a different package, but rules and aspects may only declare new files in the package of the respective target.

aehlig:
A proper fix has been pushed as https://github.com/JetBrains/hirschgarten/commit/c9752e06e8ea2f88eeb204c54ef9db43fa450f2d

PR #393, opened 2026-05-08
Support target-scoped partial sync

abbas-square:
## Why
The target resync action already creates `PartialProjectSync`, but the sync pipeline still treated model application and target persistence as full-sync-only. This makes target refresh either unsupported or too broad.

## What
- Fetch specific targets for partial sync instead of reusing a cached full resolved workspace
- Add an opt-in partial-sync hook contract and merge refreshed targets into `TargetUtils`
- Apply partial workspace model changes by replacing matching target modules while preserving project-wide entities
- Persist target dependency/generator metadata so executable target mappings can be rebuilt after partial merges
- Keep project-wide late importer work on full syncs only and update partial-sync cache tests

## Risk Assessment
Medium — this touches core sync plumbing, but the behavioral change is limited to the existing partial-sync path; full sync continues to replace all Bazel-owned workspace model entities.

## References
- Alternative implementation for #392
- `git -c core.fsmonitor=false diff --check`
- `bazel build //intellij.bazel.commons:commons`
- `bazel build //intellij.bazel.importer:importer //intellij.bazel.core:core //python/intellij.bazel.python.common:bazel-python-common` currently stops in existing importer/generated-code errors (`instrumentation`, `forEachExtensionSafeInline`, `WorkspaceFileSetExclusionCondition`)

Generated with Codex

### DIFF
--- a/intellij.bazel.commons/src/org/jetbrains/bazel/sync/environment/BazelTargetPersistenceLayer.kt
+++ b/intellij.bazel.commons/src/org/jetbrains/bazel/sync/environment/BazelTargetPersistenceLayer.kt
@@ -3,7 +3,6 @@ package org.jetbrains.bazel.sync.environment
 import com.intellij.openapi.project.Project
 import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.bazel.label.Label
-import org.jetbrains.bsp.protocol.LibraryItem
 import org.jetbrains.bsp.protocol.RawBuildTarget
 import java.nio.file.Path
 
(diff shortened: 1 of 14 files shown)


jastice:
Hi @abbas-square, thanks a lot for the PR! We're currently in the middle of rewriting the whole sync process, so I'm afraid we may not be able to merge this. @Krystian20857 can weigh in here

abbas-square:
> Hi @abbas-square, thanks a lot for the PR! We're currently in the middle of rewriting the whole sync process, so I'm afraid we may not be able to merge this. @Krystian20857 can weigh in here

@jastice understood. I'm just eager to get this feature landed in the new bazel plugin, because my entire workflow revolves around being able to search our monorepo (the entirety indexed in intellij) then only syncing what I need for the project that I'm working on. I've tried to move to the new bazel plugin a few times, but I keep going back to the old one because my entire workflow breaks if I can't have our monorepo indexed.

PR #394, opened 2026-05-11
fix: index java_proto_library classes in IDE

tboernerapex:
This should be merged after https://github.com/JetBrains/hirschgarten/pull/391

## Summary

`java_proto_library` rules apply a proto toolchain configuration transition, so their class jars are built in a `-ST-<hash>` output directory (e.g. `darwin_arm64-fastbuild-ST-8c97a2cbca67/bin`) rather than the default `fastbuild/bin`. This caused three related issues — fixed here together.

### Fix 1: Populate `java_common.jars` for `java_proto_library` from `full_compile_jars` (aspect)

`java_proto_library` targets have empty `JavaInfo.java_outputs` because the actual compilation (`bazel_java_proto_aspect`) runs in the transitioned config and its outputs are not surfaced through `java_outputs`. As a result, `java_common.jars` was always empty for these targets, and `JavaLanguagePlugin` created no library entries — so proto-generated Java classes never appeared in the IDE classpath.

**Fix**: In `java_info.bzl.template`, when `jars` is empty and the rule kind is `java_proto_library`, fall back to `JavaInfo.full_compile_jars` filtered to class jars owned by the current package only. The package filter is essential — `full_compile_jars` is transitive and without filtering every `java_proto_library` would register its entire transitive dependency tree, flooding the project model with hundreds of duplicate library entries.

```python
if not jars and provider and ctx.rule.kind == "java_proto_library":
    full_compile = [
        f for f in getattr(provider, "full_compile_jars", depset()).to_list()
        if not f.basename.endswith("-hjar.jar") and f.owner.package == ctx.label.package
    ]
```

### Fix 2: Prefer jar-bearing `TargetInfo` nodes in conflict resolution (`TargetInfoReader`)

When the BSP aspect runs alongside `bazel_java_proto_aspect`, both produce `.bsp-info.textproto` files for the same label — one in the default config (jar-less) and one in the transitioned config (contains the speed jars). The previous conflict resolution always picked the smallest-by-size node, which was the empty shell, leaving the library without any jars.

**Fix**: Among JVM nodes, prefer those with non-empty `jarsList` before falling back to smallest-by-size. The resolution logic is extracted into a testable `companion object` function with unit tests.

### Fix 3: Place materialized jdeps sibling to source file (`get_jdeps`)

`ctx.actions.declare_file("materialized_" + basename)` places the output in the package root, causing action conflicts when two targets in the same BUILD package share a basename (e.g. `TypesTest.jdeps`). Adding `sibling = raw_jdeps_file` places each materialized file next to the source jdeps in the target's own subdirectory.

### Fix 4: Retain original path for non-existent output files (`DefaultBazelOutputFileHardLinks`)

The `?: continue` that silently dropped files that don't exist on disk was changed to retain the original path, so the library entry stays registered even if the jar hasn't been downloaded. IntelliJ's VFS surfaces access errors rather than silently losing the jar from the classpath.

## Test plan

- [ ] Run `TargetInfoReaderTest` — covers all conflict-resolution branches including the proto shadow-graph case
- [ ] Sync a Bazel project containing `java_proto_library` deps and verify generated Java classes (e.g. `Account.Builder`) appear in autocomplete
- [ ] Verify no duplicate library entries for proto targets (project model size unchanged or reduced)
- [ ] Verify two `java_test` targets with the same basename in a single BUILD package build without action conflicts


tboernerapex:
This improves code completion somewhat (basically never showing before) but I still have to type 5-6 characters before I get a suggestion.

tboernerapex:
Still working on more fixes.

tboernerapex:
Well at this point i'm not sure what else to try. Some of the early changes on this PR did help the indexing but the last few commits seem to have done nothing. I was asked to repro in https://youtrack.jetbrains.com/issue/BAZEL-3169 but the example project isn't working (see the comments). With the first few changes in this PR I am now able to do fuzzy searching on the protobuf fields but nothing appears initially after you add a `.` after the variable name.

tboernerapex:
Ctrl+Space shows autocomplete just fine so the only issue is that it doesn't autocomplete after just a `.`

tboernerapex:
Unable to reproduce the issue.

PR #396, opened 2026-05-12
Fix Specs2 class filters for rules_scala tests

dkBrazz:
Import rules_scala suite metadata and use it to rewrite class-level Specs2 gutter filters into the JUnit description shape expected by the Specs2 runner Treat scala_junit_test targets as Scala test targets

~~I bundle jackson-module-kotlin with the connector so TestXmlParser resolves the Kotlin Jackson module from the same plugin classloader as the bundled Jackson XML mapper This looks like a classpath workaround, but it is required for the IDE workflow: after the Specs2 filter is fixed, Bazel passes the test and IDEA otherwise fails while parsing the produced test.xml~~

#BAZEL-3171


dkBrazz:
Dropped my XML workaround in the PR as it is addressed in BAZEL-3028 #

dkBrazz:
No longer relevant
specs2 support was improved in Scala Plugin in scope of SCL-25391

PR #397, opened 2026-06-11
Set program arguments in BazelGoRunHandler configuration

thokowick:
When debugging a Bazel go_binary target, the program arguments configured in the
run configuration were silently dropped. The debugged binary started with no
arguments, regardless of what was entered in the "Program arguments" field.

Root cause

GoRunWithDebugCommandLineState.patchAdditionalConfigs() applied the configured
environment variables to the GoApplicationConfiguration but never copied the
program arguments over, so the run config's programArguments never reached the
binary launched under dlv.

I've manually validated the behaviour, and now the arguments are properly set

### DIFF
--- a/golang/intellij.bazel.golang.common/src/debug/BazelGoRunHandler.kt
+++ b/golang/intellij.bazel.golang.common/src/debug/BazelGoRunHandler.kt
@@ -88,6 +88,9 @@ internal class GoRunWithDebugCommandLineState(
         customEnvironment[env.key] = env.value
       }
       isPassParentEnvironment = envVarsData.isPassParentEnvs
+      settings.programArguments?.let {
+        params = it
+      }
     }
   }
 }


roelofjan-elsinga:
I've given this a try and it also works for me

thokowick:
To get it to work locally with my goland installation, I also had to apply this:
```
Index: intellij.bazel.plugin/resources/META-INF/plugin.xml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/intellij.bazel.plugin/resources/META-INF/plugin.xml b/intellij.bazel.plugin/resources/META-INF/plugin.xml
--- a/intellij.bazel.plugin/resources/META-INF/plugin.xml	(revision 91fb959e22525272d42eeed4c2f355fa84835bc9)
+++ b/intellij.bazel.plugin/resources/META-INF/plugin.xml	(date 1781163823799)
@@ -53,7 +53,6 @@
     <!-- python modules -->
     <module name="intellij.bazel.python.common"/>
     <module name="intellij.bazel.python.common.performancePlugin"/>
-    <module name="intellij.bazel.python.backend"/>
 
     <!-- misc modules -->
     <module name="intellij.bazel.bytecodeViewer"/>
```

Otherwise it wouldn't load the plugin that I build with Bazel. However I didn't push this since i'm not sure that this was required for this fix

gottagofaster236 (approved):
Thanks a lot for your PR! I will add a test for it and will merge internally shortly

gottagofaster236:
Sorry for the delay, it's now merged to the upcoming 262 branch [here](https://github.com/JetBrains/hirschgarten/commit/c3f65401485c6a09e44d37a739a865e5b2dce86b) along with the [test](https://github.com/JetBrains/hirschgarten/commit/9bd7fd8419c780f7fdac884693da85a870a8fde3). We're switching the default branch from 261 from 262 soon and we're going to be releasing new plugin versions for IDEA 2026.2 from 262.
Thank you again for your pull request!

PR #399, opened 2026-06-20
Fix "Go to Definition" for exported `java_library` targets

fmeum:
## Problem

When a target's `.jdeps` references a jar produced by **another imported target** — e.g. a target reached only through an out-of-scope library that re-`exports` it — the importer turned that jar into a synthetic *sourceless* library. Because the producing target is **also** imported as a source module, the library shadowed its sources: go-to-definition on those symbols resolved to the decompiled `.class` instead of the `.java`.

## Fix

`jdepsLibraries` → `jdepsDependencies`: each jdeps jar is mapped back to the imported target that produces it (via the per-target interface/binary jars). If it belongs to an imported target it becomes a **module dependency** on that target instead of a sourceless library, so it resolves to the target's source module.

The edge uses a new `DependencyLabelKind.COMPILE_NON_EXPORTED`: it resolves for the consumer but is **not** re-exported to the consumer's own dependents (`DependencyBuilder` force-exports plain `COMPILE` module deps for Bazel targets; the new kind opts out). So a *dependent* that uses the producer without declaring it stays unresolved, and IntelliJ's existing "Add dependency on module" quick fix (`OrderEntryFix` → `BazelProjectModelModifier`) applies and edits the BUILD file.

## Test

Adds a workspace-import approval test project `jdeps_library_shadows_module`:
- `//…/a:A` (source module) ← `//…/mid:M` (`exports` A; an out-of-scope library at `import_depth: 0`) ← `//…/b:B`. B compiles against A through M's export, so B's `.jdeps` references A's header jar while A is not on B's transitive module path — the leak scenario.
- `//…/c:C` is a manual demo of the dependent behaviour: it uses A indirectly, so A is unresolved in C and the "Add dependency on module 'A'" quick fix applies. It deliberately does not compile (the workspace-model import does not compile it).

The approval golden is generated locally (the suite is local-only); it should show modules A, B and C, B → A as a non-exported module dependency, and no sourceless library for A's header jar.

Draft: golden not yet generated.

### DIFF
--- a/integrationTests/test/com/intellij/bazel/test/integration/WorkspaceImportApprovalTests.kt
+++ b/integrationTests/test/com/intellij/bazel/test/integration/WorkspaceImportApprovalTests.kt
@@ -59,6 +59,33 @@ internal class WorkspaceImportApprovalTests {
       ),
       ApprovalTestCase(WorkspaceImportApprovalTestCases.SimpleScalaTest, WorkspaceImportApprovalTestData("SimpleScalaTest")),
 
+      // Regression: a jdeps-derived sourceless library must not shadow an in-project source module.
+      // //src/main/com/example/a:A is an in-scope source module, but import_depth=0 demotes the
+      // re-exporting //src/main/com/example/mid:M to an out-of-scope library, so A's header jar appears
+      // in B's .jdeps without A being on B's transitive dependency path. Before the fix in
+      // JavaLanguagePlugin.jdepsDependencies this leaked A's header jar back in as a sourceless library
+      // that shadowed A's sources; the fix instead emits a non-exported module dependency B -> A. The
+      // golden must contain modules for A, B and C, B depending on A as a (non-exported) module, and NO
+      // sourceless library holding A's header jar. //src/main/com/example/c:C is a dependent of B that
+      // uses A indirectly -- a manual demo that A is unresolved in C (B does not re-export it) and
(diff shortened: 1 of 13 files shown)


fmeum:
@gottagofaster236 This fixes an issue I'm frequently seeing while working on Bazel due to its use of `exports`. This was agent-generated as I'm not very familiar with the codebase, but the idea and impl look reasonable to me. I didn't figure out how to set up a proper integration test, but the example that's part of the PR exactly reproduces the issue and works as expected with the fix.

Krystian20857:
@fmeum looks similar to [BAZEL-3316](https://youtrack.jetbrains.com/issue/BAZEL-3316)

fmeum:
@Krystian20857 That link says "not found"

Krystian20857:
made it public

fmeum:
Yeah, that may be an additional root cause for a similar phenomenon (or one really explains the other).

noxvost:
Generally I do not like the idea of substituting the non-imported target  сщтеутеы (sources, dependencies, etc) with jdeps. This will break a lot of internal functionality, starting from WorkspaceSnapshot/TargetUtils and ending with masking such targets from the list of all targets.

I would prefer the approach to create synthetic target which "shadows" the non-imported one. @Krystian20857 let's discuss?

FrankPortman:
This seems related to some aspects of https://github.com/JetBrains/hirschgarten/pull/401

PR #401, opened 2026-07-13
Fix resolution of generated and provider-forwarded code in the IDE

FrankPortman:
Companion to JetBrains/intellij-aspect#143. Fixes resolution of generated and provider-forwarded code, diagnosed and verified against a production monorepo (generated Scala/Python, scrooge thrift, provider-forwarding custom rules). Runnable repro of all problem classes: `pluginTests/testData/testProjects/generated_code_resolution`.

| Commit | Bug | Fix |
|---|---|---|
| 5e5a6a3 | `.srcjar` is not an archive file type, so `jar://...srcjar!/` sources roots are dead: decompiled view, manual attach fails too | Register `.srcjar` as ARCHIVE (legacy: `BlazeFileTypeFactory`) |
| 47c8c01 | `toJarUrlString()` writes `file://` URLs for `.srcjar`, unusable as sources roots | jar protocol for `jar`, `srcjar`, `zip` (legacy: `LibraryModifier`) |
| a6a2a07 | Jars already provided by a real library get a second sourceless jdeps library that wins resolution | Skip jars covered by known libraries |
| fd81ce7 | Python resolve index maps generated sources to hard-link copies outside all content roots; navigation opens orphan files where nothing resolves | Prefer the real output path, which the module content roots cover |
| 6b9e045 | Python and JVM importers derive the same module name for dual-language targets; the second entity write is rejected (`MutableEntityStorageImpl: symbolicId already exist`), so whichever importer runs second loses its module, nondeterministically per target and sync | Skip such targets in the Python importer, interim until BAZEL-2214. Removing the check was tested on the repro monorepo and breaks resolution randomly |
| 54f37d3 | | Demo project |

Two fixes still needed from intellij-aspect#143: `python_target_info.generated_sources` (feeds the resolve index, `CustomRuleTest` there), and dependency edge serialization for custom attributes, which BAZEL-3364's test classification relies on for provider-forwarding rules.

### DIFF
--- a/intellij.bazel.core/resources/intellij.bazel.core.xml
+++ b/intellij.bazel.core/resources/intellij.bazel.core.xml
@@ -483,6 +483,10 @@
       name="BazelQuery"
     />
 
+    <!-- Source jars produced by Bazel rules (e.g. scrooge) use the .srcjar extension;
+         without this mapping library source roots pointing at them cannot be opened. -->
+    <fileType name="ARCHIVE" extensions="srcjar"/>
+
     <lang.parserDefinition
       implementationClass="org.jetbrains.bazel.languages.bazelquery.parser.BazelQueryParserDefinition"
       language="BazelQuery"
--- a/python/backend/src/BazelPythonWorkspaceImporter.kt
+++ b/python/backend/src/BazelPythonWorkspaceImporter.kt
@@ -37,6 +37,7 @@ import com.jetbrains.python.sdk.PythonSdkType
 import com.jetbrains.python.sdk.createLocalSdkGuessingTypeByPath
 import org.jetbrains.annotations.ApiStatus
 import org.jetbrains.annotations.VisibleForTesting
+import org.jetbrains.bazel.commons.LanguageClass
 import org.jetbrains.bazel.commons.RepoMapping
 import org.jetbrains.bazel.magicmetamodel.formatAsModuleName
 import org.jetbrains.bazel.progress.TaskConsole
@@ -163,6 +164,10 @@ internal class BazelPythonWorkspaceImporter : BazelWorkspaceImporter {
     entitySource: EntitySource,
   ): WorkspaceImporterResult {
     for ((targetKey, target) in allPythonTargets) {
+      // Dual-language targets (e.g. custom rules forwarding both PyInfo and JavaInfo) are
+      // owned by the JVM importer: both importers derive the same module name, so creating
+      // a Python module here would clobber the JVM module and break JVM consumers.
+      if (target.rawBuildTarget.kind.languageClasses.any { it.languageName in JVM_LANGUAGE_NAMES }) continue
       val sourceDeps = pySourceDeps[targetKey] ?: emptyList()
       val sourceDependencyLibrary =
         calculateSourceDependencyLibrary(targetKey, snapshot.repoMapping, sourceDeps, entitySource, vfuManager)
@@ -419,6 +424,10 @@ internal class BazelPythonWorkspaceImporter : BazelWorkspaceImporter {
     moduleNameByKey[this] ?: this.formatAsModuleName(repoMapping)
 }
 
+
+// language name match to avoid a dependency on the java module, where the JVM
+// LanguageClass constants live
+private val JVM_LANGUAGE_NAMES = setOf("java", "kotlin", "scala")
 @ApiStatus.Internal
 @VisibleForTesting
 fun chooseSdkName(interpreter: PythonBinary, projectName: String): String =
--- a/java/intellij.bazel.java.sync/src/workspace/importer/BuilderUtils.kt
+++ b/java/intellij.bazel.java.sync/src/workspace/importer/BuilderUtils.kt
@@ -22,7 +22,7 @@ internal fun String.toResolvedVirtualFileUrl(virtualFileUrlManager: VirtualFileU
 }
 
 internal fun Path.toJarUrlString(): String =
-  if (extension == "jar") {
+  if (extension == "jar" || extension == "srcjar" || extension == "zip") {
     "jar://$this!/"
   } else {
     // There can be other library roots except for jars, e.g., Android resources. Use the file:// scheme then.
--- a/java/intellij.bazel.java.sync/src/workspace/importer/JvmBuildTargetResolver.kt
+++ b/java/intellij.bazel.java.sync/src/workspace/importer/JvmBuildTargetResolver.kt
@@ -394,11 +394,15 @@ class JvmBuildTargetResolver(
     val targetsToJdepsJars: Map<WorkspaceTargetKey, Set<Path>> =
       getAllJdepsDependencies(targetsToImport, libraryDependencies)
     val libraryNameToLibraryValueMap = HashMap<WorkspaceTargetKey, LibraryItem>()
+    // Jars already provided by a known library must not get a second, sourceless
+    // jdeps library: it would compete with the proper library during resolution.
+    val jarsCoveredByLibraries: Set<Path> =
+      libraryDependencies.values.asSequence().flatten().flatMap { it.allJars }.toSet()
     return targetsToJdepsJars.mapValues { target: Map.Entry<WorkspaceTargetKey, Set<Path>> ->
       val interfacesAndBinariesFromTarget =
         interfacesAndBinariesFromTargetsToImport.getOrDefault(target.key, emptySet())
       target.value
-        .filter { it !in interfacesAndBinariesFromTarget }
+        .filter { it !in interfacesAndBinariesFromTarget && it !in jarsCoveredByLibraries }
         .mapNotNull {
           // the synthetic label was precomputed by the plugin (it needs `bazelBin`, only available at map time)
           val label = jdepsLabelByJar[it] ?: return@mapNotNull null
--- a/pluginTests/testData/testProjects/generated_code_resolution/.bazelproject
+++ b/pluginTests/testData/testProjects/generated_code_resolution/.bazelproject
@@ -1,0 +1,3 @@
+derive_targets_from_directories: true
+directories: .
+import_depth: -1
--- a/pluginTests/testData/testProjects/generated_code_resolution/MODULE.bazel
+++ b/pluginTests/testData/testProjects/generated_code_resolution/MODULE.bazel
@@ -1,0 +1,14 @@
+bazel_dep(name = "rules_java", version = "8.16.1")
+bazel_dep(name = "rules_python", version = "1.4.0")
+bazel_dep(name = "rules_scala", version = "7.1.6")
+
+scala_config = use_extension("@rules_scala//scala/extensions:config.bzl", "scala_config")
+scala_config.settings(scala_version = "2.13.17")
+
+scala_deps = use_extension("@rules_scala//scala/extensions:deps.bzl", "scala_deps")
+scala_deps.scala()
+scala_deps.scalatest()
+scala_deps.twitter_scrooge()
+
+python = use_extension("@rules_python//python/extensions:python.bzl", "python")
+python.toolchain(python_version = "3.11")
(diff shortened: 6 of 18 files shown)


noxvost:
We need a demo project which highlights the problems being fixed. Though the changes are looking reasonable (with some flaws), it is impossible for me to evaluate the PR without seeing the root cause

FrankPortman:
@noxvost please see the demo project linked in this PR. 

I tried to make it as little code as possible, while showcasing the exact setup we have in our monorepo. I am not sure if some of what happened in my monorepo is a "perfect storm" scenario, but you can see the README (LLM generated, but I verified) in the demo project which shows the 4 things that were broken in our monorepo. Really it's 1 thing (Scala/Python codegen did not resolve) but I discovered stuff along the way to trying to fix that at varying levels.

To restate them plainly:

* Scala/Python targets didn't resolve at all
* Naive/simple fixes to the above would not let us GoToRef *inside* a code generated source to a different code generated source. This needed extra handling.
* `thrift_library` + `scrooge_scala_library` from open source `rules_scala` did not GoToRef initially, and naive/simple fixes related to the above points led to them "resolving" but not actually displaying source code via GoToRef. This needed extra handling.
* (This one I understand the least) Some of our generated Scala was only resolving from test targets. This one took the most amount of testing and iteration to fix but it seems fine with the current state of this PR + https://github.com/JetBrains/intellij-aspect/pull/143

java/intellij.bazel.java.sync/src/workspace/importer/JvmWorkspaceTargetMerger.kt
```diff
@@ -41,10 +57,12 @@ internal object JvmWorkspaceTargetMerger {
 
         // dependencies might be composed of multiple providers, so merge manually
         dependencies = (rawLeft.dependencies + rawRight.dependencies)
-          .distinctBy { it.copy(targetKey = it.targetKey.copy(aspectIds = WorkspaceAspectIds.EMPTY)) },
+          .distinctBy { it.copy(targetKey = it.targetKey.identity()) },
```
Krystian20857: Why are you stripping both configuration and aspects from dependency target key here? 
The main goal of merger is to combine targets produced by different aspects (mostly code generators) by label + configuration partial key. It's being done for following reasons: aspect provide partial key for dependencies and to avoid unnecessary module graph expansion.

So, by doing `identity()` on dependency key you're stripping too much, then you're forcing `generatedSources` to always come from "primary configuration" which is well over simplification and shouldn't be used.

java/intellij.bazel.java.sync/src/workspace/importer/JvmTargetEntitiesBuilder.kt
```diff
@@ -108,17 +109,33 @@ class ImportContext(
 
   val knownModuleNames: Set<String> = this.targets.asSequence()
     .filter { it.rawBuildTarget.kind.isJvmTarget() }
-    .mapNotNull { moduleNamesByKey[it.targetKey.copy(aspectIds = WorkspaceAspectIds.EMPTY)] }
+    .mapNotNull { moduleNamesByKey[it.targetKey.identity()] }
```
Krystian20857: This one completely breaks transition support for our plugin, basically in case of single target having two distinct configuration, plugin will try to create distinct modules for them based on configuration, after this change there will be just single module that will get random configuration ("primary configuration" from earlier)

java/intellij.bazel.java.sync/src/workspace/importer/JvmTargetEntitiesBuilder.kt
```diff
@@ -108,17 +109,33 @@ class ImportContext(
...
-        target.findBuildData<JvmBuildTarget>()?.outputJars.orEmpty()
-          .asSequence().map { it to target.targetKey }
+        // binaryOutputs and not outputJars: the latter is filtered to files that exist
+        // on disk when the hard links are created, so jars that were never built (e.g.
+        // in a configuration produced by a transition) would silently drop out of the
+        // index and the corresponding libraries would never be redirected to their
+        // source module. Identity comes from the aspect info and does not require the
+        // files to exist.
+        (target.findBuildData<JvmBuildTarget>()?.binaryOutputs ?: SourceFileCollection.EMPTY).getFiles()
+          .map { it.normalizeOutputPath() to target.targetKey }
```
Krystian20857: This one was suppose to substitute compiled JARs (mainly hjars) which are pulled by manually constructed `JavaInfo` provider by code generator.

java/intellij.bazel.java.sync/src/workspace/importer/TestTargetClassifier.kt
```diff
@@ -4,45 +4,65 @@ import org.jetbrains.annotations.ApiStatus
...
+
+    // Everything any non-test executable depends on, transitively, is production code and
+    // must never be marked as test sources. ExecutableTargetsComputer cannot be used to
+    // answer this: its same-package attribution shortcut and its result cap are run
+    // configuration heuristics, and both can hide production dependents of a library that
+    // also has a test right next to it (e.g. a code generation macro emitting a library
+    // together with a same-package sourceless test).
+    val dependenciesOf = targets.associate { it.id to it.dependencies.map { dep -> dep.targetKey.label } }
+    fun reachableFrom(roots: Collection<Label>): Set<Label> {
+      val visited = HashSet<Label>(roots)
```
Krystian20857: @mrkocot Can you take a look at this?

intellij.bazel.core/resources/intellij.bazel.core.xml
```diff
@@ -478,6 +478,10 @@
       name="BazelQuery"
     />
 
+    <!-- Source jars produced by Bazel rules (e.g. scrooge) use the .srcjar extension;
+         without this mapping library source roots pointing at them cannot be opened. -->
+    <fileType name="ARCHIVE" extensions="srcjar"/>
+
```
Krystian20857, line 489: Great! I was having similar issue myself some time ago, will add this one.

python/backend/src/BazelPythonWorkspaceImporter.kt
```diff
@@ -158,6 +158,10 @@ internal class BazelPythonWorkspaceImporter : BazelWorkspaceImporter {
     entitySource: EntitySource,
   ): WorkspaceImporterResult {
     for ((target, _) in snapshot.allTargets.filterBuildTarget<PythonBuildTarget>()) {
+      // Dual-language targets (e.g. custom rules forwarding both PyInfo and JavaInfo) are
+      // owned by the JVM importer: both importers derive the same module name, so creating
+      // a Python module here would clobber the JVM module and break JVM consumers.
+      if (target.rawBuildTarget.kind.isJvmTarget()) continue
```
Krystian20857: This is a known issue, however we should aim for more generic solution
https://youtrack.jetbrains.com/issues/BAZEL?preview=BAZEL-2214

Krystian20857 (commented):
Thanks for PR,
However I'll take a look at provided reproduction cases myself and check what's exactly is going on inside.

java/intellij.bazel.java.sync/src/workspace/importer/JvmWorkspaceTargetMerger.kt
FrankPortman: In relation to this and your other comment - it is possible this is no longer needed. I think/hope this was related to the most tedious use case to test/squash, which is that some transitioned targets only resolved from test files. It's possible this is no longer needed.

Let me do some tests and get back to you, hopefully it's not needed since I see your point about configuration/transitions being significant.

java/intellij.bazel.java.sync/src/workspace/importer/TestTargetClassifier.kt
FrankPortman: i think this ended up being the fix for what https://github.com/JetBrains/hirschgarten/pull/401#discussion_r3588881318 was also related to. I am hoping the configuration stuff was not actually needed. will check

python/backend/src/BazelPythonWorkspaceImporter.kt
FrankPortman: that link doesn't seem to take me to any specific issue

java/intellij.bazel.java.sync/src/workspace/importer/JvmWorkspaceTargetMerger.kt
FrankPortman: image

tested on our monorepo without the config merge commit: everything still resolves, including the transitioned targets. 

the test classification fix was the actual fix for that it seems.

the state of this PR now reflects this.

java/intellij.bazel.java.sync/src/workspace/importer/JvmTargetEntitiesBuilder.kt
FrankPortman: from what I understand, the problem with `outputJars` for this index is the existence filter, i.e. in a transitioned configuration the jars are never built locally, so exactly those entries drop out and the libraries never redirect to their source modules. 

`binaryOutputs` carries the same identities without needing files on disk. if `outputJars` has to stay as the hjar-substituted view for the codegen case, happy to point the index at a dedicated field instead.

Krystian20857: Fair point, I'll take closer look at this particular case

python/backend/src/BazelPythonWorkspaceImporter.kt
Krystian20857: ah, issue is private, but related to java + python module names over lapping, we're working on this one

java/intellij.bazel.java.sync/src/workspace/importer/TestTargetClassifier.kt
mrkocot: I've tested it with both normal and tricky cases I remember used to be problematic, works for them well

python/backend/src/BazelPythonWorkspaceImporter.kt
Krystian20857: @FrankPortman could you please check whenever removing this check impact code resolution inside your repository?

FrankPortman: tested it: unfortunately this breaks resolution. claude claims it breaks non-deterministically based on a race condition (the workspace model rejects the second write for dual-language targets), but I did not try a bunch of cold runs because of how long it takes to sync my full monorepo.

Verbatim from the log:

```
 MutableEntityStorageImpl - Entity with symbolicId: ModuleId(name=<redacted>-514860d) already exist
```

In this case many objects were fine but I happened to see a few breaking in different test files that were fine before. Confirmed it's normal again after adding this commit back.

One extra data point, that Claude seems to think is relevant:  even with this change I still see `LibraryId` collisions between the `*_python` entities of configuration duplicates, so the overlap isn't just java vs python.

Verbatim from Claude:

```
Yes, exactly. On the current review-tip build (skip in place), this session's log shows:

- ModuleId collisions: 0 — the java-vs-python module clobber is fully gone
- LibraryId collisions: 126 — all shaped like <pkg>.<target>_python-514860d already exist, i.e. the Python importer's own source-dependency libraries colliding with themselves: two instances of the same Python target (config/aspect duplicates) both writing a library under the same name

It's benign in practice — the colliding libraries are content-identical duplicates, so whichever write wins carries the same roots, which is why everything stays green. But it demonstrates the naming overlap is python-vs-python too, not just java-vs-python — useful scope information for their generic BAZEL-2214 fix, which is why it earns the one-sentence mention in the comment.
```

java/intellij.bazel.java.sync/src/workspace/importer/TestTargetClassifier.kt
FrankPortman: can i resolve this comment?

intellij.bazel.core/resources/intellij.bazel.core.xml
FrankPortman: thanks!

java/intellij.bazel.java.sync/src/workspace/importer/JvmTargetEntitiesBuilder.kt
FrankPortman: reverted since it doesn't seem to be required anymore

PR #403, opened 2026-08-04
BAZEL-2087: rerun a single test or the failed tests from the results tree with any test runner

jlnhws:
## Problem

Re-running tests from the test results tree doesn't work unless `use_jetbrains_test_runner` is enabled — see [BAZEL-2087](https://youtrack.jetbrains.com/issue/BAZEL-2087):

- Right-clicking a test node offers "Nothing here": `BazelRerunTestConfigurationProducer` returns early for every project where the flag is off.
- The "rerun failed tests" toolbar action isn't even attached to the console in that case.

Both features rely on re-running by the JetBrains runner's test unique ids (the `JB_TEST_UNIQUE_IDS` environment variable), which no other runner understands.

## Change

Add a generic path that works with any test runner, built on a bazel `--test_filter` (a standard flag honored by the native JUnit runner and custom runners alike) instead of `JB_TEST_UNIQUE_IDS`.

**Commit 1 — rerun a single test (context menu):**
- New per-language extension point `org.jetbrains.bazel.testFilterProvider` (`BazelTestFilterProvider`) that turns a selected test node's location URL into a `--test_filter`.
- Java implementation (`JavaTestFilterProvider`) that reads back a `java:test://…` / `java:suite://…` location hint and emits the same filter the Java gutter "run test" action does, so re-running from the gutter and from the context menu behave identically.
- To read those hints back, the delimiters in `BazelJavaTestLocationHintProvider` (`java:test`, `://`, `$`, `/`) change from `private` to `internal`. Both files are in `intellij.bazel.java.core`, so the side that writes a location hint and the side that parses it share one definition of the format rather than spelling it out twice.
- `BazelRerunTestConfigurationProducer` branches on the flag: JetBrains runner → existing unique-ids path; otherwise → derive a `--test_filter` and `setTestFilter`.

**Commit 2 — rerun failed tests (toolbar):**
- Attach the rerun-failed action for every test run, not only when the JetBrains runner is used.
- `BazelRerunFailedTestsAction` builds a `--test_filter` (an alternation of the failed tests) via the same provider. Only **leaf** tests contribute — `getFailedTests` also returns the defective parent classes/containers, and turning one into a filter (e.g. a bare `FooTest`) would re-run the whole class rather than just the failures.

The provider is an extension point, so other languages (Kotlin, Go, …) can add their own formatters; this PR implements Java.

## Testing

- `JavaTestFilterProviderTest` — location-URL → `--test_filter` mapping (single method, whole suite, nested class, test-name anchoring, non-Java URLs).
- `BazelRerunFailedTestsActionTest` — `failedTestsToFilter`: leaf-only selection (the defective parent class is dropped), alternation of the failed tests, and `null` when only containers failed.
- Manually verified in IntelliJ against a custom (non-JetBrains) JUnit 5 runner with `use_jetbrains_test_runner` **off**: right-clicking a single test, a test inside a `@Nested` class, or a whole test class re-runs exactly that test/class (e.g. `--test_filter=com.example.FooTest.NestedTest.it_passes$`); filtering for `it_fails` does not also run `it_fails_again`; and the rerun-failed toolbar button re-runs only the failing leaf tests, with no whole-class entries in the filter.

### DIFF
--- a/intellij.bazel.core/resources/intellij.bazel.core.xml
+++ b/intellij.bazel.core/resources/intellij.bazel.core.xml
@@ -43,6 +43,10 @@
                     interface="org.jetbrains.bazel.run.test.BazelTestLocatorProvider"
                     dynamic="true"
     />
+    <extensionPoint qualifiedName="org.jetbrains.bazel.testFilterProvider"
+                    interface="org.jetbrains.bazel.run.test.BazelTestFilterProvider"
+                    dynamic="true"
+    />
     <extensionPoint qualifiedName="org.jetbrains.bazel.bzlmodResolver"
                     interface="org.jetbrains.bazel.languages.starlark.bazel.bzlmod.BazelModuleResolver"
                     dynamic="true"
--- a/intellij.bazel.core/src/org/jetbrains/bazel/run/BazelCommandLineStateBase.kt
+++ b/intellij.bazel.core/src/org/jetbrains/bazel/run/BazelCommandLineStateBase.kt
@@ -105,11 +105,12 @@ abstract class BazelCommandLineStateBase(environment: ExecutionEnvironment) : Co
     val actions = createActions(console, handler, executor)
 
     val executionResult = DefaultExecutionResult(console, handler, *actions)
-    if (useJetBrainsTestRunner) {
-      val rerunFailedTestsAction = BazelRerunFailedTestsAction(console)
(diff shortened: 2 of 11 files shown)


gottagofaster236 (approved):
Thx a lot for the PR! I imported it with minor changes and will merge it after approval from my colleagues

PR #404, opened 2026-08-04
BAZEL-2087: rerun failed tests from the toolbar with any test runner

jlnhws:
## Problem

The "rerun failed tests" toolbar action in the test results is only usable when `use_jetbrains_test_runner` is enabled: it's attached to the test console only in that case, and `BazelRerunFailedTestsAction` re-runs by the JetBrains runner's test unique ids (`JB_TEST_UNIQUE_IDS`), which no other runner understands.

## Change

- Attach the rerun-failed action for **every** test run in `BazelCommandLineStateBase`, not only when the JetBrains runner is used.
- `BazelRerunFailedTestsAction` now branches on the flag: JetBrains runner → existing unique-ids path; otherwise → build a generic bazel `--test_filter` (an alternation of the failed tests) via `BazelTestFilterProvider` and `setTestFilter`.
- Only **leaf** tests contribute to the filter. `getFailedTests` also returns the defective parent classes/containers, and turning one of those into a filter (e.g. a bare `FooTest`) would re-run the whole class — every test in it — rather than just the failures.

## Testing

- Added `BazelRerunFailedTestsActionTest`, covering `failedTestsToFilter`: leaf-only selection (the defective parent class is dropped), alternation of the failed methods, and `null` when only non-leaf containers failed.
- Manually verified in IntelliJ with `use_jetbrains_test_runner` off: after a run with failures, the rerun-failed button re-runs only the failing leaf tests (`--test_filter=Foo.a$|Foo.b$…`) with no whole-class entries.

### DIFF
--- a/intellij.bazel.core/resources/intellij.bazel.core.xml
+++ b/intellij.bazel.core/resources/intellij.bazel.core.xml
@@ -43,6 +43,10 @@
                     interface="org.jetbrains.bazel.run.test.BazelTestLocatorProvider"
                     dynamic="true"
     />
+    <extensionPoint qualifiedName="org.jetbrains.bazel.testFilterProvider"
+                    interface="org.jetbrains.bazel.run.test.BazelTestFilterProvider"
+                    dynamic="true"
+    />
     <extensionPoint qualifiedName="org.jetbrains.bazel.bzlmodResolver"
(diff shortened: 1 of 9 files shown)


jlnhws:
Combined into #403 at the maintainer's request (two commits: single-test rerun, then failed-tests rerun).

PR #405, opened 2026-08-05
Fail the sync immediately when a required project view import is missing

jlnhws:
## Problem

When a project view uses a required `import` (e.g. `.bazelproject` = `import local.bazelproject`) and the imported file does **not** exist, the import is silently dropped and the resulting project view is effectively empty. `ReparseProjectViewFilePreSyncHook` reports the missing import as an error diagnostic — but the sync doesn't stop. It proceeds into the pipeline and **contacts the Bazel server** to resolve configurations/targets over the empty project view.

That work is doomed, and instead of surfacing the real cause it fails slowly with confusing *secondary* errors:
- `ProjectResolver — bazel config invocation failed, falling back to BEP configurations`
- `targetInfos should not be empty`
- `ProjectResolver — Serializer for class 'BazelConfiguration' is not found`

…taking tens of seconds before it finally fails. 

## Fix

Add a pre-sync gate in `ProjectSyncTask`: after the project view is reparsed, if it still has unresolved **required** imports, abort the sync (`FAILURE`) **before contacting the Bazel server**. The offending `import` line was already reported with its file/position by the pre-sync hook, so the user sees the real cause immediately. The check is factored into `ProjectView.unresolvedRequiredImports()`, shared by the hook (reporting) and the gate (aborting).

Result: instead of a doomed server sync with misleading secondary errors, the sync stops near-instantly at the missing-import error.

## Testing

- Added `ProjectViewUnresolvedImportsTest` for `unresolvedRequiredImports()` (required-but-missing is returned; `try_import`-style optional is not).
- Verified in IntelliJ with `.bazelproject` = `import local.bazelproject` and `local.bazelproject` deleted (idea.log timestamps):
  - **Before:** sync reaches the server — `bazel config invocation failed`, `targetInfos should not be empty`, `Serializer for class 'BazelConfiguration' is not found` — and fails after ~30s.
  - **After:** sync aborts at the pre-sync gate immediately, no server contact, surfacing the missing-import error.

### DIFF
--- a/intellij.bazel.backend/src/org/jetbrains/bazel/sync/task/ProjectSyncTask.kt
+++ b/intellij.bazel.backend/src/org/jetbrains/bazel/sync/task/ProjectSyncTask.kt
@@ -28,6 +28,8 @@ import org.jetbrains.bazel.config.rootDir
 import org.jetbrains.bazel.coroutines.BazelCoroutineService
 import org.jetbrains.bazel.fus.BazelSyncCollector
 import org.jetbrains.bazel.label.Label
+import org.jetbrains.bazel.languages.projectview.ProjectViewService
+import org.jetbrains.bazel.languages.projectview.unresolvedRequiredImports
 import org.jetbrains.bazel.performance.bspTracer
 import org.jetbrains.bazel.progress.syncConsole
 import org.jetbrains.bazel.progress.withSubtask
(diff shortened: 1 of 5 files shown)


aehlig (approved):
LGTM. Thanks.

Will import.

noxvost:
the changes are landed into master/262

PR #408, opened 2026-09-17
Add missing backend dependency to intellij.bazel.java.common

fmeum:
`JavaBazelRules` implements `TargetKindProvider` from `intellij.bazel.backend` since BAZEL-3081, and the module's `.iml` already depends on it, but the Bazel BUILD file was not regenerated. This makes the open source build fail on `262`:

```
java/intellij.bazel.java.common/src/sync/workspace/languages/jvm/JavaBazelRules.kt:7:43: error: unresolved reference 'targetKind'.
import org.jetbrains.bazel.sync.workspace.targetKind.TargetKindProvider
java/intellij.bazel.java.common/src/sync/workspace/languages/jvm/JavaBazelRules.kt:9:32: error: unresolved reference 'TargetKindProvider'.
internal class JavaBazelRules: TargetKindProvider {
```

Adding the dependency to the `bazel-java-common` target, matching the `.iml`, makes `bazel build //:plugin-bazel_zip` succeed again.


fmeum:
@aehlig

gottagofaster236:
Thanks a lot! Merging this now. We were going to implement this synchronization to be automatic a while ago, probably now is a good time to do it finally

gottagofaster236:
Merged

PR #409, opened 2026-09-17
Inject Shell Script into genrule cmd strings

fmeum:
Stacked on #408, which fixes the `262` build; the first commit here is that fix.

Starlark string literals now implement `PsiLanguageInjectionHost`, with a `LiteralTextEscaper` that decodes Starlark escape sequences (and treats raw strings literally). The existing `ElementManipulator` escapes edited content back; a raw string literal stays raw as long as the new content can be represented in it and is otherwise converted into a regular literal with the same value. A `MultiHostInjector` injects the Shell Script language into the `cmd` and `cmd_bash` arguments of `genrule` / `native.genrule` calls, so these strings get shell highlighting, completion, inspections, ShellCheck and "Edit fragment" support when the bundled Shell Script plugin is available.

The argument value can be a single (possibly triple-quoted) literal, a `+` concatenation, a `"sep".join([...])` call on a list literal, or parentheses around any of these. All literals end up in one injected file; other expressions are represented by a `missing_value` placeholder.

Also makes the string contents range account for `r`/`b` prefixes and unterminated strings.

The Shell Script plugin is added to the bundled plugins used in tests and to the `pluginTests` module dependencies.


fmeum:
This isn't ready yet, I'm still iterating on it. Will mark rfr when done.

intellij.bazel.core/src/org/jetbrains/bazel/languages/starlark/injection/GenruleCommandShellInjector.kt
```diff
@@ -0,0 +1,161 @@
...
+    literal.createLiteralTextEscaper().decode(literal.getStringContentsOffset(), decoded)
+    return decoded.toString()
+  }
+
+  private fun isGenruleCommand(value: PsiElement): Boolean {
+    val argument = value.parent as? StarlarkNamedArgumentExpression ?: return false
+    val argumentName = argument.name ?: return false
+    if (argumentName !in SHELL_COMMAND_ARGUMENTS) return false
+    val call = (argument.parent as? StarlarkArgumentList)?.parent as? StarlarkCallExpression ?: return false
+    // Not StarlarkCallExpression.getName(), which is the value of the `name` attribute for rule targets.
```
fmeum:
```suggestion
```

intellij.bazel.core/src/org/jetbrains/bazel/languages/starlark/utils/StarlarkQuote.kt
```diff
@@ -12,13 +12,17 @@ enum class StarlarkQuote(val quote: String) {
...
+    get() = this == TRIPLE_SINGLE || this == TRIPLE_DOUBLE
+
   fun rangeWithinQuotes(string: String): TextRange = TextRange(quote.length, string.length - quote.length)
 
   fun wrap(toWrap: String): String = quote + toWrap + quote
 
   fun unwrap(toUnwrap: String): String = toUnwrap.removeSurrounding(quote)
 
   companion object {
+    // Triple quotes have to be checked first, as a triple quote also starts with a single quote.
```
fmeum:
```suggestion
```

